PR #14, opened 2019-04-22
[compiler] Support operators with multiple outputs

bwasti:
* #22 [compiler] Improve conversion from PyTorch IR to Relay
* #21 [testing] Improve testing to include models, image inputs and autotuned parameters
* **#14 [compiler] Support operators with multiple outputs**

### DIFF
--- a/torch_tvm/compiler.cpp
+++ b/torch_tvm/compiler.cpp
@@ -129,15 +129,28 @@ tvm::relay::Function TVMCompiler::convertToRelay(
     frontier = new_frontier;
   }
 
-  AT_ASSERT(subgraph->outputs().size() == 1);
-  auto output = subgraph->outputs().at(0);
-  AT_ASSERT(value_map.find(output) != value_map.end());
-  tvm::Array<tvm::relay::Var> free_vars =
-      tvm::relay::FreeVars(value_map[output]);
-  AT_ASSERT(free_vars.size() == input_vars.size());
+  auto output = ([&]() -> tvm::relay::Expr {
+    if (subgraph->outputs().size() == 1) {
+      auto sg_output = subgraph->outputs().at(0);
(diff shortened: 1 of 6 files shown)


torch_tvm/compiler.cpp
```diff
@@ -108,24 +108,49 @@ tvm::relay::Function TVMCompiler::convertToRelay(
...
+      return value_map[sg_output];
+    }
+    tvm::NodePtr<tvm::relay::TupleNode> n = tvm::make_node<tvm::relay::TupleNode>();
+    tvm::Array<tvm::relay::Expr> fields;
+    for (const auto& sg_output : subgraph->outputs()) {
+      AT_ASSERT(value_map.find(sg_output) != value_map.end());
+      fields.push_back(value_map[sg_output]);
+    }
+    n->fields = std::move(fields);
+    return tvm::relay::Tuple(n);
```
wanchaol: can we actually make even the single output also fills into a Tuple, so that we could merge the single and multiple output case together? that way we could avoid checking the output size everytime.

torch_tvm/compiler.cpp
```diff
@@ -108,24 +108,42 @@ tvm::relay::Function TVMCompiler::convertToRelay(
         if (use.user->outputs().size() < 1) {
           continue;
         }
-        // TODO handle multiple outputs
-        AT_ASSERT(use.user->outputs().size() == 1);
-        value_map[use.user->output()] = getOperator(use.user, relay_inputs);
-        new_frontier.emplace_back(use.user->output());
+        // if there are 2+ outputs, getOperator returns a tuple
+        if (use.user->outputs().size() == 1) {
```
wanchaol: shouldn't this if can be combine together with the else statement if in the batchnorm we eliminate the conditional on output==1?

wanchaol (approved):
LGTM, one minor suggestion

PR #16, opened 2019-04-22
[testing] Refactor to use hypothesis style testing

bwasti:
* #19 WIP
* #18 [models] WIP
* #15 [compiler] Free vars may be less than input vars (some inputs ignored)
* **#16 [testing] Refactor to use hypothesis style testing**

### DIFF
--- a/test/test_operators.py
+++ b/test/test_operators.py
@@ -1,74 +1,137 @@
 import unittest
-from test.util import TestCase
-
+from test.util import TVMTest
+from hypothesis import given, settings, strategies as st
 import torch
 import torch.nn.functional as F
 import torch
 
 # test jit tvm operators
-class TestOperators(TestCase):
+class TestOperators(TVMTest):
+    @TVMTest.given(shape=TVMTest.rand_shape(rank=1))
+    def test_add(self, shape):
+        x = torch.rand(shape)
+        y = torch.rand(shape)
+        z = torch.rand(shape)
 
-    def test_add(self):
         def add(a, b, c):
-          return a + b + c
+            return a + b + c
 
-        self.checkTraceTVM(add, verbose=True)
+        ref_out, tvm_out = self.runBoth(add, x, y, z)
+        assert torch.allclose(ref_out, tvm_out)
 
-    def test_mul(self):
+    @TVMTest.given(shape=TVMTest.rand_shape(rank=1))
+    def test_mul(self, shape):
+        x = torch.rand(shape)
+        y = torch.rand(shape)
+        z = torch.rand(shape)
+
         def mul(a, b, c):
             return a * b * c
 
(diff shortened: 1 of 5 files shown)


test/test_operators.py
```diff
@@ -1,74 +1,137 @@
...
+        bias = torch.rand(num_kernels)
+
+        def conv(a, b):
+            return F.conv2d(a + a, b)
+
+        def conv_bias(a, b, c):
+            return F.conv2d(a + a, b, c)
+
+        ref_out, tvm_out = self.runBoth(conv, X, W)
+        assert torch.allclose(ref_out, tvm_out, rtol=0.01, atol=0.01)
```
wanchaol: I think it would be better to standardize the rtol and atol instead of calling them each time, probably make a wrapper with `torch.allclose` and make `rtol` and `atol` be defaults.

test/util.py
```diff
@@ -6,129 +6,80 @@
...
+    def rand_shape(
+        cls, min_rank=1, max_rank=4, min_dim=1, max_dim=2 ** 4, like=None, rank=None
+    ):
+        def get():
+            rank_ = rank
+            if not rank_:
+                rank_ = cls.rand_int(min_rank, max_rank)()
+            return cls.rand_list(cls.rand_int(min_dim, max_dim), rank_)()
+
+        return get
```
wanchaol: shall we also make a rand_tensor? I think it would be beneficial so that we don't need to construct the tensor everyone we wrote the test.

wanchaol (approved):
looks great! have some nits I think it's good to have, but it's good to go 👍

test/test_operators.py
```diff
@@ -1,74 +1,137 @@
...
+    @TVMTest.given(
+        shape=TVMTest.rand_shape(rank=4, min_dim=4, max_dim=4),
+        kernel_size=TVMTest.rand_int(3, 3),
+        num_kernels=TVMTest.rand_int(5, 5),
+    )
+    def test_conv_simple(self, shape, kernel_size, num_kernels):
+        # NCHW
+        X = torch.rand(shape)
+        W = torch.rand((num_kernels, shape[1], kernel_size, kernel_size))
+        bias = torch.rand(num_kernels)
```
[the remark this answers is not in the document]
bwasti: what do you mean? like put torch.rand inside the given()?  certainly a good idea :)  perhaps I can do that in a follow up diff

test/test_operators.py
bwasti: it's not a huge amount to copy and paste over. I like avoiding wrappers in test cases so that anyone new to the project can copy paste and immediately tweak what they need

test/util.py
bwasti: follow up diff? :)

PR #20, opened 2019-04-29
[operator] Add maxpool 

wanchaol:
maxpool2d is added in this PR. There're some unstableness when `ceil_mode=True`, which generate mismatched output shapes between pytorch and tvm.

### DIFF
--- a/test/test_operators.py
+++ b/test/test_operators.py
@@ -1,6 +1,7 @@
 import unittest
 from test.util import TVMTest
 from hypothesis import given, settings, strategies as st
+
 import torch
 import torch.nn.functional as F
 import torch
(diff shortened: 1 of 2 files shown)


test/test_operators.py
```diff
@@ -133,5 +134,30 @@ def avg_pool2d_strides_ceil_mode_pad(a):
...
+        X = torch.rand(shape)
+        def max_pool2d(a):
+            return F.max_pool2d(a, 3) + 2.0
+
+        def max_pool2d_strides_padding_ceil_mode(a):
+            return F.max_pool2d(
+                a, 2, stride=stride, padding=1, ceil_mode=False
+            )
+
+        # TODO: fix the unstableness when ceil_mode=True case
```
bddppq: To support ceil_mode you can adjust the padding value to pass to tvm.

wanchaol: Oh in fact relay do have a `ceil_mode`, and their ways of handling padding seems to be same with what pytorch did (but it turned out different which needs some debugging), so i am not sure if adjust the padding value here will help.

test/test_operators.py
```diff
@@ -1,6 +1,7 @@
 import unittest
-from test.util import TVMTest
+from util import TVMTest
```
bwasti: pls don't change this :)

try using `python -m unittest test.test_operators.TestOperators.test_max_pool2d`

wanchaol: oops didn't mean to commit this, it's only for my local testing, I will remove

PR #17, opened 2019-04-22
[compiler] Add None type handling with sentinel value

bwasti:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* #24 [WIP] Use PackedFunc to enable custom TVM operator registration from python
* #23 [benchmark] Add initial benchmark file
* #22 [compiler] Improve conversion from PyTorch IR to Relay
* #21 [testing] Improve testing to include models, image inputs and autotuned parameters
* **#17 [compiler] Add None type handling with sentinel value**
* #14 [compiler] Support operators with multiple outputs

PR #21, opened 2019-05-02
[testing] Improve testing to include models, image inputs and autotuned parameters

bwasti:
* #22 [compiler] Improve conversion from PyTorch IR to Relay
* **#21 [testing] Improve testing to include models, image inputs and autotuned parameters**


test/util.py
```diff
@@ -60,26 +61,33 @@ def f_impl(*args, **kwargs):
...
-
-        return ref_out, tvm_out
+        with torch.no_grad():
+          # jit the function
+          trace_jit = torch.jit.trace(func, inputs)
+          ref_out = trace_jit(*inputs)
+
+          # jit the function and lower to TVM
+          torch_tvm.enable()
+          with autotvm.apply_history_best("test/autotvm_tuning.log"):
```
wanchaol: this actually will not be that transparent to the user. I think we might can make the `torch_tvm.enable` API take if we do autotvm and pass the log there, in the future we will need to think about how much tunable tvm api that we want to expose to the user.

test/test_models.py
```diff
@@ -329,25 +329,119 @@ def resnext101_32x8d(pretrained=False, **kwargs):
...
+        image = io.imread('test/cat.png')[:,:,:3].transpose(2,0,1)
+        input_image = torch.unsqueeze(torch.Tensor(image), 0)
+        ref_out, tvm_out = self.runBoth(model, input_image)
+        # With CNN model tests we look for top-k similarity rather than
+        # exact numerical matches post-softmax.
+        k = 10
+        torch.testing.assert_allclose(ref_out.topk(k).indices, tvm_out.topk(k).indices)
+
+    def test_resnet18(self):
+        self.model_test(resnet18)
```
wanchaol: curious, there should be somewhere import `resnet18` from torchvision, where i did not see those import statements though

wanchaol (approved):

test/util.py
bwasti: everything is exposed to the user by default pretty much

test/test_models.py
bwasti: it's defined above -- I just copied and pasted it in

PR #22, opened 2019-05-02
[compiler] Improve conversion from PyTorch IR to Relay

bwasti:
* **#22 [compiler] Improve conversion from PyTorch IR to Relay**
* #21 [testing] Improve testing to include models, image inputs and autotuned parameters


bwasti:
This PR predominantly fixes bugs in conversion

torch_tvm/operators.cpp
```diff
@@ -186,6 +217,25 @@ RegisterTVMOperator reg({
        auto out = tvm::relay::CallNode::make(op, inputs, tvm::Attrs(), {});
        return out;
      }},
+    {Symbol::fromQualString("aten::threshold_"),
```
wanchaol: are you getting `aten::threshold_` from the trace rather than `aten::relu`?

torch_tvm/version.py
```diff
@@ -0,0 +1,9 @@
+# This file is generated by setup.py. DO NOT EDIT!
```
wanchaol: what's this used for? can we add it to gitignore instead?

torch_tvm/operators.cpp
```diff
@@ -186,6 +217,25 @@ RegisterTVMOperator reg({
...
+     [](Node* node, tvm::Array<tvm::relay::Expr> inputs) {
+       AT_CHECK(!relayIsNone(inputs[0]));
+       AT_CHECK(!relayIsNone(inputs[1]));
+       AT_CHECK(!relayIsNone(inputs[2]));
+       auto d = relayToConstant<float>(inputs[1]);
+       AT_CHECK(d <  1e-7, "aten::threshold_ only supported for threshold 0, got", d);
+       AT_CHECK(d > -1e-7, "aten::threshold_ only supported for threshold 0, got", d);
+       AT_CHECK(d <  1e-7, "aten::threshold_ only supported for threshold 0, got", d);
```
wanchaol: it looks like there are 2 duplicate checks here

torch_tvm/operators.cpp
```diff
@@ -186,6 +217,25 @@ RegisterTVMOperator reg({
...
+       AT_CHECK(d > -1e-7, "aten::threshold_ only supported for threshold 0, got", d);
+       d = relayToConstant<float>(inputs[2]);
+       AT_CHECK(d <  1e-7, "aten::threshold_ only supported for value 0, got", d);
+       AT_CHECK(d > -1e-7, "aten::threshold_ only supported for value 0, got", d);
+       AT_CHECK(d <  1e-7, "aten::threshold_ only supported for value 0, got", d);
```
wanchaol: same here

torch_tvm/operators.cpp
wanchaol: `aten::threshold` is too generic to pass to tvm, I think we should only consider `aten::relu`

torch_tvm/compiler.cpp
```diff
@@ -23,24 +28,33 @@ tvm::relay::Var TVMCompiler::convertToRelay(const Value* val) {
 }
 
 tvm::relay::Expr TVMCompiler::convertToRelay(const IValue& val) {
+  // All doubles are converted to floats
```
wanchaol: looks like we can convert to `float64` for double, can you add a comment here why the conversion to `float32` is necessary?

torch_tvm/compiler.cpp
```diff
@@ -69,14 +85,15 @@ tvm::relay::Expr TVMCompiler::convertToRelay(const IValue& val) {
 }
 
 tvm::relay::Function TVMCompiler::convertToRelay(
-    std::shared_ptr<Graph> subgraph) {
+    std::shared_ptr<Graph> subgraph, std::vector<Value*>* input_values) {
```
wanchaol: just take a reference of the vector as argument?

torch_tvm/compiler.cpp
```diff
@@ -199,11 +224,21 @@ void TVMCompiler::run(Stack& stack) {
...
+  for (auto i = 0; i < input_values_.size(); ++i) {//subgraph_->inputs().size(); ++i) {
+    //auto* ivalue = value_to_ivalue[subgraph_->inputs()[i]];
+    auto* value = input_values_[i];
+    if (!value_to_ivalue.count(value)) {
+      auto optional_ivalue = toIValue(value);
+      AT_ASSERT(optional_ivalue.has_value());
+      value_to_ivalue[value] = optional_ivalue.value();
+    }
+    auto ivalue = value_to_ivalue.at(input_values_[i]);
+    auto tensor = ivalue.toTensor().to(caffe2::TypeMeta::Make<float>());
```
wanchaol: I think we can do something like `ivalue.toTensor().to(at::kFloat)`

torch_tvm/version.py
bwasti: ah this shouldn't even be in this commit

torch_tvm/operators.cpp
bwasti: I'm getting threshold_ from the trace, not relu

torch_tvm/compiler.cpp
bwasti: it's not really necessary, I'll swap to fp64

torch_tvm/compiler.cpp
bwasti: input_values is being written to, so I'm passing it as a pointer

PR #33, opened 2019-05-10
[hotfix] Disable dilation test in conv (which is broken in TVM during auto-layout)

bwasti:
* **#33 [hotfix] Disable dilation test in conv (which is broken in TVM during auto-layout)**


bwasti:
If CI is green I think this is worth landing :)

PR #36, opened 2019-05-12
[testing] Add the test for group conv

llyfacebook:
There is another PR to experimental TVM to make it work with layout xform and without layout xform.


bwasti:
this will need to also update the TVM submodule to point to the newer commit, right?
https://github.com/facebookexperimental/tvm/commit/2008fa44bf0ec1336c5e2b518d615dc9852716af

llyfacebook:
> this will need to also update the TVM submodule to point to the newer commit, right?
> [facebookexperimental/tvm@2008fa4](https://github.com/facebookexperimental/tvm/commit/2008fa44bf0ec1336c5e2b518d615dc9852716af)

Actually, it is okay, since it can still use the default schedule.

PR #37, opened 2019-05-14
[compiler] naive fallback to jit interpreter

wanchaol:
* **#37 [compiler] naive fallback to jit interpreter**


bwasti (changes requested):
Can you dump the exception in the warning?
Also, can you expose a "strict" mode and add that to tests?

bwasti (approved):

PR #41, opened 2019-05-15
[cleanup] Cleaning up the default values for TVMCompiler

bwasti:
* **#41 [cleanup] Cleaning up the default values for TVMCompiler**

### DIFF
--- a/torch_tvm/compiler.h
+++ b/torch_tvm/compiler.h
@@ -14,13 +14,18 @@ struct TVMObject {
   tvm::PackedFunc get_output;
 };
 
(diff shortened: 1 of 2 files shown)


torch_tvm/compiler.h
```diff
@@ -14,13 +14,18 @@ struct TVMObject {
   tvm::PackedFunc get_output;
 };
 
+#define PT_TVM_DEFAULT_OPT_LEVEL 3
```
yinghai: Why not constexpr stuff?

PR #42, opened 2019-05-20
[compiler][api] Enable custom registration of Relay ops from python

bwasti:
* **#42 [compiler][api] Enable custom registration of Relay ops from python**

### DIFF
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -32,7 +32,7 @@ class IntegrationTest(TVMTest):
         assert torch.allclose(o1, o2, rtol=0.01, atol=0.01)
 
         # Ensure JIT and TVM JIT both work
-        ref_out, tvm_out = self.runBoth(reshape, X, X)
+        ref_out, tvm_out = self.runBoth(test, X, X)
         assert torch.allclose(ref_out, tvm_out, rtol=0.01, atol=0.01)
 
 
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -18,15 +18,23 @@ class IntegrationTest(TVMTest):
           return d + e
         
         X = torch.randn(128,128)
-        print(test.graph)
-        print(test.graph_for(X,X))
-        o = test(X, X)
+        o1 = test(X, X)
 
+        torch_tvm.disable()
+
+        # Ensure eager mode works
         def test(a, b):
           d = torch.ops.tvm.woo(a)
           e = torch.ops.tvm.woo(b)
           return d + e
-        o = test(X, X)
+
+        o2 = test(X, X)
+        assert torch.allclose(o1, o2, rtol=0.01, atol=0.01)
+
+        # Ensure JIT and TVM JIT both work
+        ref_out, tvm_out = self.runBoth(reshape, X, X)
+        assert torch.allclose(ref_out, tvm_out, rtol=0.01, atol=0.01)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/test/test_integration.py
+++ b/test/test_integration.py
@@ -1,0 +1,32 @@
+import unittest
+from test.util import TVMTest
+import torch_tvm
+import torch
+import tvm
+from tvm import relay
+
+class IntegrationTest(TVMTest):
+    def test_basic(self):
+        torch_tvm.register_operator("woo", "nn.relu")
+
+        torch_tvm.enable()
+
+        @torch.jit.script
+        def test(a, b):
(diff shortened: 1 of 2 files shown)


torch_tvm/register.cpp
```diff
@@ -22,55 +23,93 @@ static std::string device_type = "cpu";
...
+            auto op = tvm::relay::Op::Get(relay_name);
+            AT_ASSERT(op);
+            return tvm::relay::CallNode::make(op, inputs, tvm::Attrs(), {});
+          }}});
+
+    auto relay_op = tvm::relay::Op::Get(relay_name);
+    AT_ASSERT(relay_op);
+    std::vector<Argument> torch_inputs;
+    Graph wrapper_graph;
+    std::vector<Value *> graph_inputs;
```
wanchaol: nit: `Value*`?

torch_tvm/register.cpp
```diff
@@ -22,55 +23,93 @@ static std::string device_type = "cpu";
...
+    RegisterTVMOperator reg_tvm_operator(
+        {{Symbol::fromQualString("tvm::" + torch_name),
+          [relay_name](Node *node, tvm::Array<tvm::relay::Expr> inputs) {
+            auto op = tvm::relay::Op::Get(relay_name);
+            AT_ASSERT(op);
+            return tvm::relay::CallNode::make(op, inputs, tvm::Attrs(), {});
+          }}});
+
+    auto relay_op = tvm::relay::Op::Get(relay_name);
+    AT_ASSERT(relay_op);
```
wanchaol: this should be `AT_CHECK` and send msg to user if there's no such op exist?

torch_tvm/register.cpp
```diff
@@ -22,55 +23,93 @@ static std::string device_type = "cpu";
...
 
   m.def("disable", []() { fusion_enabled = false; });
+  m.def("register_operator", [tvm_sym](std::string torch_name,
+                                       std::string relay_name) {
+    RegisterTVMOperator reg_tvm_operator(
+        {{Symbol::fromQualString("tvm::" + torch_name),
+          [relay_name](Node *node, tvm::Array<tvm::relay::Expr> inputs) {
+            auto op = tvm::relay::Op::Get(relay_name);
+            AT_ASSERT(op);
+            return tvm::relay::CallNode::make(op, inputs, tvm::Attrs(), {});
```
wanchaol: does this mean that we currently can only register simple operators that does not require custom Attrs?

torch_tvm/register.cpp
```diff
@@ -22,55 +23,93 @@ static std::string device_type = "cpu";
...
+    auto sym = Symbol::fromQualString("tvm::" + torch_name);
+    Node *node = wrapper_graph.create(sym, graph_inputs);
+    wrapper_graph.appendNode(node);
+    AT_CHECK(node->outputs().size() == 1,
+             "Currently only single output relay ops "
+             "can be custom registered from Python.");
+    wrapper_graph.registerOutput(node->output());
+    // Convert node to a tvm compilation group containing the node
+    node = SubgraphUtils::createSingletonSubgraph(node, tvm_sym);
+    auto cc = std::make_shared<TVMCompiler>(node, opt_level, strict,
```
wanchaol: from what I observed here, for each custom operators, we are creating a single separate `TVMCompiler` group and execute them separately? will the `customFusion` pass take the compilation group and merge it with the bigger one?

test/test_integration.py
```diff
@@ -0,0 +1,32 @@
...
+
+        @torch.jit.script
+        def test(a, b):
+          d = torch.ops.tvm.woo(a)
+          e = torch.ops.tvm.woo(b)
+          return d + e
+        
+        X = torch.randn(128,128)
+        print(test.graph)
+        print(test.graph_for(X,X))
```
wanchaol: it is hard to see the graph from the PR, why don't we just call `runBoth` here?

torch_tvm/register.cpp
bwasti: yep, unfortunately.  There isn't a great way to expose the relay Attrs interface to python (as far as I can tell)

torch_tvm/register.cpp
bwasti: customFusion will never be run on these operators because this interface is for eager mode only

wanchaol (approved):
looks good.

bwasti:
There are some issues with this API that make it pretty useless, so I am going to abandon it.  That being said, there are some tricks here worth using

PR #43, opened 2019-05-23
[compiler] Expose relay ops registered as torch.ops.tvm.__

bwasti:
* #44 [compiler] Expose `to_relay` API which hands the Relay graph over in Python
* **#43 [compiler] Expose relay ops registered as torch.ops.tvm.__**

### DIFF
--- a/torch_tvm/compiler.cpp
+++ b/torch_tvm/compiler.cpp
@@ -270,12 +270,8 @@ void TVMCompiler::run(Stack& stack) {
     tvm::runtime::NDArray ret_val = cache_[spec].get_output(i);
     auto dl_tensor = ret_val.ToDLPack();
     auto tensor = at::fromDLPack(dl_tensor);
-    auto new_tensor = at::empty_like(tensor);
-    auto new_dl_tensor = at::toDLPack(new_tensor);
-    ret_val = cache_[spec].get_output(i, &new_dl_tensor->dl_tensor);
-    auto var = torch::autograd::make_variable(at::fromDLPack(new_dl_tensor));
+    auto var = torch::autograd::make_variable(tensor);
     stack.push_back(IValue(var));
-
     i++;
   }
 }
--- a/torch_tvm/compiler.cpp
+++ b/torch_tvm/compiler.cpp
@@ -190,6 +190,9 @@ TVMCompiler::TVMCompiler(
   }
   ctx_.device_id = 0;
   subgraph_ = node->g(attr::Subgraph);
+  auto pfb = tvm::runtime::Registry::Get("relay.build_module._BuildModule");
+  AT_ASSERT(pfb);
+  build_mod_ = (*pfb)();
 }
 
 void TVMCompiler::run(Stack& stack) {
@@ -204,7 +207,6 @@ void TVMCompiler::run(Stack& stack) {
 
   CompleteArgumentSpec spec{false, ArrayRef<IValue>(inputs)};
 
-  std::vector<Value*> input_values_;
   if (cache_.find(spec) == cache_.end()) {
     for (auto& kv : value_to_ivalue) {
       kv.first->inferTypeFrom(kv.second.toTensor());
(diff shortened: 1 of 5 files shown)


bwasti:
There is currently a bug lurking here regarding the lifetime management of DLManagedTensors.  TVM stores NDArrays in the graph runtime and on program exit will attempt to delete them, assuming that it has ownership of the DLManagedTensors' tensors.  This ends up breaking everything and the program segfaults.  A small patch to get around this is

```
diff --git a/src/runtime/ndarray.cc b/src/runtime/ndarray.cc
index 39c17b8b..92747b1b 100644
--- a/src/runtime/ndarray.cc
+++ b/src/runtime/ndarray.cc
@@ -190,8 +190,10 @@ void NDArray::CopyFromTo(DLTensor* from,
 using namespace tvm::runtime;

 void NDArrayDLPackDeleter(DLManagedTensor* tensor) {
-  static_cast<NDArray::Container*>(tensor->manager_ctx)->DecRef();
-  delete tensor;
+  if (tensor) {
+    static_cast<NDArray::Container*>(tensor->manager_ctx)->DecRef();
+    delete tensor;
+  }
 }

 int TVMArrayAlloc(const tvm_index_t* shape,
```

but realistically we should revisit the ownership model and how values are passed in to TVM

torch_tvm/operators.cpp
```diff
@@ -14,12 +20,50 @@ std::unordered_map<Symbol, TVMOpFunctor>& getTVMOperatorMap() {
...
+        wrapper_graphs.emplace_back();
+        auto &wrapper_graph = wrapper_graphs.back();
+        std::vector<Value *> torch_inputs;
+        for (const auto &inp : schema.arguments()) {
+          torch_inputs.emplace_back(wrapper_graph.addInput());
+        }
+        Node* node =
+            wrapper_graph.create(op.sym, torch_inputs, schema.returns().size());
+        wrapper_graph.appendNode(node);
+        wrapper_graph.registerOutput(node->output());
```
wanchaol: this is a bit confusing. `wrapper_graph` create and append the node into its graph, then it does not ever get used when we creating the `SingletonSubgraph`?

torch_tvm/operators.cpp
```diff
@@ -14,12 +20,50 @@ std::unordered_map<Symbol, TVMOpFunctor>& getTVMOperatorMap() {
...
+
+        Symbol tvm_sym = Symbol::fromQualString("tvm::CompilationGroup");
+        node = SubgraphUtils::createSingletonSubgraph(node, tvm_sym);
+        auto cc = std::make_shared<TVMCompiler>(node);
+
+        // NB: We assume all relay ops are pure
+        auto options = OperatorOptions().aliasAnalysis(AliasAnalysisKind::PURE);
+        // We don't know the true number of inputs so we make it varargs
```
wanchaol: we do know the true number of arguments right? even if in runtime, the number of inputs are the same as `schema.arguments()` I believe.

torch_tvm/compiler.cpp
```diff
@@ -214,7 +216,7 @@ void TVMCompiler::run(Stack& stack) {
     // or fall back to the JIT interpreter for execution
     tvm::relay::Function tvm_func;
     try {
-      tvm_func = convertToRelay(subgraph_, &input_values_);
+      tvm_func = convertToRelay(subgraph_, &cache_[spec].input_values);
```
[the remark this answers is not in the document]
bwasti: C++ unordered_map's [] creates an object if it doesn't exist

torch_tvm/operators.cpp
bwasti: it's a lifetime management hack

torch_tvm/operators.cpp
bwasti: oh yea we do! dated comment

wanchaol (approved):
looks good

PR #44, opened 2019-05-23
[compiler] Expose `to_relay` API which hands the Relay graph over in Python

bwasti:
* **#44 [compiler] Expose `to_relay` API which hands the Relay graph over in Python**
* #43 [compiler] Expose relay ops registered as torch.ops.tvm.__

### DIFF
--- a/torch_tvm/compiler.cpp
+++ b/torch_tvm/compiler.cpp
@@ -8,7 +8,7 @@
 
 using namespace torch::jit;
 
-tvm::relay::Var TVMCompiler::convertToRelay(Value* val) {
+tvm::relay::Var TVMCompiler::convertToRelay(Value* val, TVMContext ctx) {
   auto optional_ivalue = toIValue(val);
   if (optional_ivalue.has_value()) {
     val->inferTypeFrom(optional_ivalue.value().toTensor());
@@ -27,11 +27,11 @@ tvm::relay::Var TVMCompiler::convertToRelay(Value* val) {
   AT_ASSERT(0);
 }
 
-tvm::relay::Expr TVMCompiler::convertToRelay(const IValue& val) {
+tvm::relay::Expr TVMCompiler::convertToRelay(const IValue& val, TVMContext ctx) {
   // All doubles are converted to floats
   if (val.isDouble()) {
     auto x = tvm::runtime::NDArray::Empty(
-        {}, tvm::runtime::String2TVMType("float32"), ctx_);
+        {}, tvm::runtime::String2TVMType("float32"), ctx);
     auto d = val.toDouble();
     AT_CHECK(d <= std::numeric_limits<float>::max());
     AT_CHECK(d >= std::numeric_limits<float>::lowest());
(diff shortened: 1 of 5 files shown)


torch_tvm/register.cpp
```diff
@@ -72,5 +75,33 @@ PYBIND11_MODULE(_torch_tvm, m) {
 
   m.def("disable", []() { fusion_enabled = false; });
 
+  m.def("push_relay_expr", [](std::shared_ptr<Graph> g,
```
wanchaol: since we don't expose the `push_relay_expr` as a public API, should we make it as `_push_relay_expr`?

torch_tvm/register.cpp
```diff
@@ -21,11 +23,12 @@ static int opt_level = 2;
 static std::string device_type = "cpu";
 static std::string device = "llvm -mcpu=core-avx2";
 static std::string host = "llvm -mcpu=core-avx2";
+static auto tvm_sym = Symbol::fromQualString("tvm::CompilationGroup");
 
+static std::unordered_map<size_t, tvm::relay::Expr> relay_exprs;
+static size_t relay_exprs_uuid = 1337;
```
wanchaol: what is this number specially for? I think it can also starts at 0?

torch_tvm/compiler.cpp
```diff
@@ -27,11 +27,11 @@ tvm::relay::Var TVMCompiler::convertToRelay(Value* val) {
   AT_ASSERT(0);
 }
 
-tvm::relay::Expr TVMCompiler::convertToRelay(const IValue& val) {
+tvm::relay::Expr TVMCompiler::convertToRelay(const IValue& val, TVMContext ctx) {
   // All doubles are converted to floats
   if (val.isDouble()) {
     auto x = tvm::runtime::NDArray::Empty(
-        {}, tvm::runtime::String2TVMType("float32"), ctx_);
```
wanchaol: I am a bit surprised this could access the`ctx_` before..

torch_tvm/register.cpp
```diff
@@ -72,5 +75,33 @@ PYBIND11_MODULE(_torch_tvm, m) {
...
+        AT_CHECK(subgraph->inputs().size() == inputs.size(), "Expected ", subgraph->inputs().size(), " inputs");
+        for (auto i = 0; i < inputs.size(); ++i) {
+          subgraph->inputs()[i]->inferTypeFrom(inputs[i]);
+        }
+        TVMContext ctx;
+        ctx.device_type = kDLCPU;
+        ctx.device_id = 0;
+        auto expr = TVMCompiler::convertToRelay(subgraph, ctx);
+        relay_exprs[++relay_exprs_uuid] = expr;
+        return relay_exprs_uuid;
```
wanchaol: that will only convert the first subgraph to relay right? Shall we give the user warning here indicating they could not convert the full graph to relay, they will need to add more operators to make sure a single subgraph

torch_tvm/register.cpp
```diff
@@ -1,9 +1,11 @@
 #include <pybind11/pybind11.h>
+#include <stack>
```
wanchaol: I did not see where this header is needed?

wanchaol (approved):
looks nice! some nits

torch_tvm/compiler.cpp
bwasti: it's a trick! this wasn't a static method before

PR #58, opened 2019-07-10
add fuse admm/matmul back to linear pass

wanchaol:
* **#58 add fuse admm/matmul back to linear pass**
* #57 [codemod] deprecated macros


torch_tvm/fuse_linear.cpp
```diff
@@ -0,0 +1,50 @@
...
+    std::string addmm_pattern = R"IR(
+    graph(%input, %weight, %bias, %4):
+        %weight_t = aten::t(%weight)
+        %res = aten::addmm(%bias, %input, %weight_t, %4, %4)
+        return (%res))IR";
+    std::string matmul_add_pattern = R"IR(
+    graph(%input, %weight, %bias, %4):
+        %weight_t = aten::t(%weight)
+        %output = aten::matmul(%input, %weight_t)
+        %res = aten::add_(%output, %bias, %4)
```
bwasti: should we also match on non-inplace add?

test/test_operators.py
```diff
@@ -206,9 +206,15 @@ def test_linear(self, shape, out_features):
         def linear(input, weight, bias):
             return F.linear(input, weight, bias) + 2.0
 
+        def linear_no_bias(input, weight):
```
bwasti: maybe we should test the pattern matching in python?

bwasti (approved):
nice! :)

test/test_operators.py
wanchaol: Yeah sure we will need to pybind the pass to python, I will add a test

torch_tvm/fuse_linear.cpp
wanchaol: We can but I don't think we need, people using `nn.Linear` will always get this inplace `add` through tracing/scripting, outplace `add` will be matched if people manually write `matmul + outplace add`.

test/test_operators.py
bwasti: can we avoid pybind by manually constructing a torchscript model containing t + matmul + _add and just ensuring "linear" exists in the graph_for call?

PR #62, opened 2019-07-18
[fusion] Migrate away from CustomFuseGraph

bwasti:
* #63 [WIP] Loop impl
* **#62 [fusion] Migrate away from CustomFuseGraph**
* #59 [operators][fix] Prevent segfault if the passed in tensor hasn't been dimensioned yet


bwasti:
This diff is a hand-rolled version of customFuseGraph that can handle control flow and avoids some of the older code that would likely be more effort to refactor

Depends on the change made to pytorch here:
https://github.com/pytorch/pytorch/pull/23005

PR #68, opened 2019-07-18
[WIP] Switch to using ProfiledTensorType

Krovatkin:


torch_tvm/compiler.cpp
```diff
@@ -10,24 +10,36 @@ using namespace torch::jit;
...
 tvm::relay::Var TVMCompiler::convertToRelay(Value* val, TVMContext ctx) {
   auto optional_ivalue = toIValue(val);
-  if (optional_ivalue.has_value()) {
+
+  tvm::Array<HalideIR::Expr> sizes;
+  if (auto ptt = val->type()->cast<ProfiledTensorType>())
+  {
+
+    auto csizes = ptt->sizes().concrete_sizes();
+    TORCH_INTERNAL_ASSERT(csizes.has_value());
```
bwasti: should we handle the case when this isn't true?

torch_tvm/compiler.cpp
```diff
@@ -222,29 +233,40 @@ void TVMCompiler::run(Stack& stack) {
...
     // either throw or fall back to the JIT interpreter for execution
+    cache_ = TVMObject {};
     tvm::relay::Function tvm_func;
     try {
-      tvm_func = convertToRelay(subgraph_, ctx_, &cache_[spec].input_values);
+      tvm_func = convertToRelay(subgraph_, ctx_, &(*cache_).input_values);
+      // we compiled the subgraph successfully
+      (*cache_).invalid = false;
     } catch (const std::exception& e) {
+      (*cache_).invalid = true;
```
wanchaol: When we failing back to JIT, this means some operators are not converted successfully due to operator semantic mismatch and other behaviors, this invalid flag will let the compiler to re-run the conversion everytime for the same inputs and it will always fail, so I am not sure if that flag would be necessary.

torch_tvm/compiler.cpp
```diff
@@ -222,29 +233,40 @@ void TVMCompiler::run(Stack& stack) {
...
 
-  CompleteArgumentSpec spec{false, ArrayRef<IValue>(inputs)};
-
-  if (cache_.find(spec) == cache_.end()) {
+  if (!cache_ || (cache_ && (*cache_).invalid)) {
     for (auto& kv : value_to_ivalue) {
-      kv.first->inferTypeFrom(kv.second.toTensor());
+      // TODO: convince Fuser to NOT create TVMCompilationGroups
+      // if ANY of subgraph inputs weren't profiled
+      TORCH_INTERNAL_ASSERT(kv.first->type()->cast<ProfiledTensorType>());
```
bwasti: I don't think we need this, as it is done on line 111

torch_tvm/compiler.cpp
```diff
@@ -222,29 +233,40 @@ void TVMCompiler::run(Stack& stack) {
...
     }
     // bail out mechanism: try to convert to Relay, if it fails to convert the
     // graph by any reason(i.e. op difference), depend on the user preference,
     // either throw or fall back to the JIT interpreter for execution
+    cache_ = TVMObject {};
```
bwasti: at this point, if we run a graph twice with different sizes will we still get TVM compiled code each time?
is the logic for that moved up into the profiled executor?

torch_tvm/compiler.cpp
```diff
@@ -222,29 +233,40 @@ void TVMCompiler::run(Stack& stack) {
...
             "Pytorch TVM: fail to convert to relay, exception: ", e.what());
       }
+
       LOG(WARNING)
-          << "Pytorch TVM: fail to convert to relay, falling back to JIT for execution, exception: "
+          << "Pytorch TVM: fail to convert to relay, exception: "
           << e.what() << "\n";
+    }
+
+    if ((*cache_).invalid)
```
bwasti: can this block be merged back into block on line 260?

torch_tvm/compiler.cpp
```diff
@@ -255,38 +277,41 @@ void TVMCompiler::run(Stack& stack) {
     tvm::runtime::Module mod = mod_f();
     auto pfr = tvm::runtime::Registry::Get("tvm.graph_runtime.create");
     AT_ASSERT(pfr);
+
     tvm::runtime::Module run_mod =
         (*pfr)(json, mod, (int)ctx_.device_type, (int)ctx_.device_id);
-    cache_[spec].set_input = run_mod.GetFunction("set_input_zero_copy", false);
-    cache_[spec].kernel = run_mod.GetFunction("run", false);
-    cache_[spec].get_output = run_mod.GetFunction("get_output", false);
+    (*cache_).set_input = run_mod.GetFunction("set_input_zero_copy", false);
```
bwasti: `->`?

torch_tvm/compiler.cpp
```diff
@@ -222,29 +233,40 @@ void TVMCompiler::run(Stack& stack) {
     value_to_ivalue[value_input] = inputs[i];
   }
 
-  CompleteArgumentSpec spec{false, ArrayRef<IValue>(inputs)};
-
-  if (cache_.find(spec) == cache_.end()) {
+  if (!cache_ || (cache_ && (*cache_).invalid)) {
```
bwasti: if `cache_` is an optional, can we get rid of the `invalid` attribute somehow?  it's pretty confusing

torch_tvm/register.cpp
```diff
@@ -35,6 +36,7 @@ PYBIND11_MODULE(_torch_tvm, m) {
   RegisterOperators op({Operator(
       tvm_sym,
       [](const Node* node) {
+        GRAPH_DUMP("A graph passed to TVMCompiler\n", node->g(attr::Subgraph));
```
bwasti: can we move this to a different diff?

torch_tvm/compiler.cpp
Krovatkin: yup, we will bail out, profile again and generate a graph with a TVMCompGroups for different shapes.

torch_tvm/compiler.cpp
Krovatkin: I think we shouldn't just create a TVMCompGroup in this case. Even though, we can compile at a runtime if we get a workload where shapes change all the time it will be pretty wasteful. I'm looking into changing Fuser to not fuse things that don't have `ProfiledTensorTypes`

torch_tvm/compiler.cpp
Krovatkin: arrgh, thanks! I forgot to switch it

torch_tvm/register.cpp
Krovatkin: sure

torch_tvm/compiler.cpp
bwasti: if it's alternating between 20 batch size shapes but run 1M times it's probably worth still compiling each shape

torch_tvm/compiler.cpp
Krovatkin: exactly! we don't want to re-run if we already know it's going to fail! Re-running compilations might be pretty expensive

torch_tvm/compiler.cpp
Krovatkin: hopefully, this will be handled via bailouts. We specialize for shape_set1. Then if we see another frequent set, we will specialize for that one as well and so on.

blaquezzz (approved):

Issue #55, opened 2019-06-28
Funny fusion effects - constant as output of compilation group


t-vi:
Hi,

for fusing a not that much longer elementary pointwise computation, I get
```
with tvm::CompilationGroup_0 = graph(%0 : Float(*),
      %1 : Float(*)):
  %4 : int = prim::Constant[value=1]()
  %2 : int = prim::Constant[value=1]()
  %3 : Float(*) = aten::add(%0, %1, %2)
  return (%3, %4)
```
and the second output is then used outside the group, which is not good, as it'll hit errors when fusing the next thing.

Best regards

Thomas

t-vi:
Trying to produce a minimal example, I got:
```
RuntimeError: IValue is not a Scalar
The above operation failed in interpreter, with the following stack trace:
at <ipython-input-15-e6d508d37f4e>:3:12
@torch.jit.script
def func(x)->torch.Tensor:
    return x+torch.tanh(x+x)
           ~~~~~~~~~~~~~~~~ <--- HERE
```

t-vi:
So the first error seems related to `prim::Constant` being declared fusible.

As far as I understand, `prim::Constant` nodes are not usually declared fusible but instead are pulled in (but duplicated) when they are arguments to a fused node. This happens in GraphFuser::mergeNodeIntoGroup:

https://github.com/pytorch/pytorch/blob/2132ea1d8d6ca9cb2ab6c2206c7cdc9203e1908f/torch/csrc/jit/passes/graph_fuser.cpp#L359-L365

ulyssesxxxi:
Hi @t-vi,
I'm experiencing the same issue, while trying to compile a model from maskrcnn-benchmark. Do you know if there is any workaround until the issue is fixed?
Thx.

t-vi:
Yes, sorry, I have been meaning to send a PR, but you simply need to remove `prim::Constant` fusion.
https://github.com/pytorch/tvm/blob/7d3fa3fbd88855a1655c174cc43caac7fe6b954e/torch_tvm/operators.cpp#L417

I see, however, that #72 is moving away from the CustomGraphFuser (which ironically seems to be doing funny const handling, too, when the current approach makes fixing this simple).

PR #72, opened 2019-07-23
[fusion] Migrate away from CustomFuseGraph

bwasti:
* **#72 [fusion] Migrate away from CustomFuseGraph**

### DIFF
--- a/torch_tvm/fusion_pass.cpp
+++ b/torch_tvm/fusion_pass.cpp
@@ -19,11 +19,6 @@ value_list sortReverseTopological(ArrayRef<Value*> inputs, Block* block) {
 
 bool canHandle(Block* block, AliasDb& aliasDb);
 bool canHandle(Node* node, AliasDb& aliasDb) {
-  if (aliasDb.hasWriters(node)) {
-    if (!aliasDb.isInPlace(node)) {
-      return false;
-    }
-  }
   if (node->kind() == prim::Constant) {
     return true;
   }
@@ -53,6 +48,17 @@ c10::optional<Node*> tryMerge(
       aliasDb.moveBeforeTopologicallyValid(producer, consumer);
 
   if (!canMerge) {
+    return c10::nullopt;
+  }
+
+  // Consumer is only allowed to have writers
+  if (aliasDb.hasInputWriters(consumer)) {
+    if (!aliasDb.isInPlace(producer)) {
+      return c10::nullopt;
+    }
+  }
+
+  if (aliasDb.hasOutputWriters(consumer)) {
     return c10::nullopt;
   }
 
--- a/torch_tvm/fusion_pass.cpp
+++ b/torch_tvm/fusion_pass.cpp
@@ -3,6 +3,7 @@
 
 #include <torch/csrc/jit/passes/common_subexpression_elimination.h>
 #include <torch/csrc/jit/passes/dead_code_elimination.h>
+#include <torch/csrc/jit/jit_log.h>
 
 using namespace torch::jit;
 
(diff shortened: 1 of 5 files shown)


bwasti:
this depends on https://github.com/pytorch/pytorch/pull/23210

wanchaol (commented):
I tried to use this to compare the tvm fusion group we generated before and after, and looks like we are pulling in redundant constant nodes again and again, which I think we shouldn't. Taking a example:

```
with tvm::CompilationGroup_1 = graph(%0 : Tensor,
      %1 : Float(*, *, *),
      %2 : Float(*)):
  %11 : int[] = prim::Constant[value=[0]]()
  %3 : int[] = prim::Constant[value=[1]]()
  %4 : int[] = prim::Constant[value=[0]]()
  %5 : int[] = prim::Constant[value=[1]]()
  %6 : bool = prim::Constant[value=0]()
  %7 : int[] = prim::Constant[value=[0]]()
  %8 : int = prim::Constant[value=1]()
  %9 : bool = prim::Constant[value=1]()
  %x1.1 : Tensor = aten::_convolution(%0, %1, %2, %3, %4, %5, %6, %7, %8, %6, %6, %9) # code/my_noqqq.py:377:8
  return (%x1.1)
with tvm::CompilationGroup_2 = graph(%0 : Tensor,
      %1 : Float(*, *, *),
      %2 : Float(*)):
  %11 : int[] = prim::Constant[value=[0]]()
  %3 : int[] = prim::Constant[value=[1]]()
  %4 : int[] = prim::Constant[value=[0]]()
  %5 : int[] = prim::Constant[value=[1]]()
  %6 : bool = prim::Constant[value=0]()
  %7 : int[] = prim::Constant[value=[0]]()
  %8 : int = prim::Constant[value=1]()
  %9 : bool = prim::Constant[value=1]()
  %x2.1 : Tensor = aten::_convolution(%0, %1, %2, %3, %4, %5, %6, %7, %8, %6, %6, %9) # code/my_noqqq.py:380:8
  return (%x2.1)
```
I thought that the graph executor run constant pooling somewhere but it turned out not running it properly.. this will probably need to do more passes after we generate the fusion group like the graphFuser did:

```
  // After FuseGraph some common subexpressions may come back
  EliminateCommonSubexpression(graph);
  // We might have emitted a fair amount of useless shape propagating code, so
  // remove it
  EliminateDeadCode(graph);
  // Improve the quality of shape propagation code that was left
  PeepholeOptimizeShapeExpressions(graph->block());
```

Also, we might think of just adding constants as arguments rather than pulling it into the fusion group, this way it's easier to run constant pooling stuff I think.

t-vi:
So I'm ignorant of the underlying goal of not using the graph fuser, but the constants seem to be an indication that it might be nice to try to use as much of PyTorch's capabilities as possible.

In contrast to the current master torch_tvm (see #55) and apparently also this PR, the PyTorch fuser treats constants more delicately: It will _not_ fuse them on their own but instead copies constant inputs of operation nodes to be fused into the fusion group and reconnects the fused operation to these copies.

bwasti:
@wanchaol @t-vi good points on being careful with `prim::Constant`, I will update that to be consistent with the current graph fusion code.

@t-vi The issue is that GraphFuser doesn't (and probably shouldn't) handle the more complex cases of aliasing and control flow.  For the sake of Relay lowering, we will want to have loops and conditionals as well as tensor views well supported to generate the best code.  I put up this diff as the first step in many to get that all working 😃

t-vi:
Good point about the control flow which would not be needed for basic graph fusion. I'm sure that people will come up with other uses for control-flow capable fusers if they get them... :)

torch_tvm/fusion_pass.cpp
```diff
@@ -0,0 +1,107 @@
...
+graph_node_list::iterator scanNode(
+    Node* consumer,
+    AliasDb& aliasDb,
+    Block* block) {
+  auto inputs = sortReverseTopological(consumer->inputs(), block);
+  for (auto input : inputs) {
+    if (auto group = tryMerge(consumer, input->node(), aliasDb)) {
+      // we successfully merged, so the new group's `inputs` may have
+      // changed. So rescan the new group for more merging opportunities.
+      return group.value()->reverseIterator();
```
kimishpatel: has_value check?

yinghai:
@bwasti Could you add some reasoning on this PR? Like why do we need to migrate away from custom graph fuse.

zrphercule:
Is this pr ready to be merged now?

bwasti:
@yinghai I added some notes above in response to @t-vi who had a similar question.  Largely it comes down to proper handling of alias information and enabling control flow in the future

@zrphercule I believe so, yes.  I will add a few more tests on coverage

torch_tvm/fusion_pass.h
```diff
@@ -0,0 +1,7 @@
+#include <torch/csrc/jit/passes/alias_analysis.h>
+#include <torch/csrc/jit/passes/graph_fuser.h>
+#include <torch/csrc/jit/passes/utils/subgraph_utils.h>
+
+void FuseSupportedOps(std::shared_ptr<torch::jit::Graph> graph);
```
yinghai: Actually, is this function TVM specific? If not, can we move it to pytorch repro?

bwasti: that's a good catch.  For the sake of efficiency we can "stage" a lot of simple functionality like this here and upstream it to PyTorch at some point.

I'd like to prioritize flushing out functionality before trying to standardize it into an API.  That being said, I've tried to make it very copy-and-paste-able 😛

jackm321: >very copy-and-paste-able

Lol sounds useful

yinghai: > That being said, I've tried to make it very copy-and-paste-able

That's something that can be landed directly into Pytorch in the spirit of hackability of PyTorch. Lol

torch_tvm/fusion_pass.cpp
```diff
@@ -0,0 +1,134 @@
...
+    Block* body = node->blocks().at(0);
+    return canHandle(body, aliasDb);
+  }
+  return isSupported(node);
+}
+
+bool canHandle(Block* block, AliasDb& aliasDb) {
+  for (Node* node : block->nodes()) {
+    if (!canHandle(node, aliasDb)) {
+      return false;
```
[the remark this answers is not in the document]
t-vi: The traditional fuser works on blocks, traversing them in the fusion pass. This here seems more about whether we can fuse everything in the block in order to fuse the entire block-using control flow statement.

torch_tvm/fusion_pass.cpp
t-vi: if it has not, the if will evaluate to false.

torch_tvm/fusion_pass.cpp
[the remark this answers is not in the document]
jackm321: If I understand what you mean, that's what I'm saying. It looks the the traditional fuser recurses on sub-blocks fusing what it can within them whereas this will only fuse a sub-block if all nodes within it are fusable (recursively). So I think the traditional fuser could for example fuse most of a loop body while leaving the any unsupported nodes in the loop body unfused wheras this I think will try to fuse the entire loop or nothing.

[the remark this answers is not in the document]
bwasti: yep, this means any single fusion attempt will be all or nothing.  We can add attempts to recursively fuse on blocks that weren't fused on a previous attempt (in a later diff maybe?) to recover the CustomGraphFuse behavior

PR #73, opened 2019-07-24
Customized layer norm for pytorch tvm repo.

kimishpatel:
Summary:
This PR adds layer norm topi op. Provides correponding relay op.
Adds test cases.

Test Plan:
python custom_tvm_ops/tests/layer_norm_test.py

Reviewers:

Subscribers:

Tasks:

Tags:


custom_tvm_ops/attrs/layer_norm_attrs.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include <tvm/relay/expr.h>
+#include <tvm/attrs.h>
+
+namespace tvm {
+namespace relay {
+struct CustomLayerNormAttrs : public tvm::AttrsNode<CustomLayerNormAttrs> {
+  Array<Integer> axis;
+  bool affine{false};
```
llyfacebook: Can you also put this field in the declare statement?

custom_tvm_ops/tests/layer_norm_test.py
```diff
@@ -0,0 +1,202 @@
...
+source_location = os.path.abspath(__file__)
+pytorch_tvm_location = os.path.join(source_location, "../../..")
+pytorch_tvm_location = os.path.abspath(pytorch_tvm_location)
+logger = logging.getLogger()
+lib_path=os.path.join(pytorch_tvm_location, \
+        "build/custom_tvm_ops/topi/libcustom_tvm_ops_topi.so")
+torch.ops.load_library(lib_path)
+lib_path=os.path.join(pytorch_tvm_location, \
+        "build/custom_tvm_ops/relay/libcustom_tvm_ops_relay.so")
+torch.ops.load_library(lib_path)
```
llyfacebook: I think the library can be linked into torch_tvm and installed.

CMakeLists.txt
```diff
@@ -23,10 +23,12 @@ link_directories(${PYTORCH_DIR}/lib)
 
 add_subdirectory(pybind11)
 add_subdirectory(${TVM_DIR})
+add_subdirectory(custom_tvm_ops)
```
bwasti: I think it's easier to just add a folder `torch_tvm/ops/` and add `${CMAKE_CURRENT_SOURCE_DIR}/torch_tvm/ops/*.cpp` after line 4.  This will avoid the need for new linked libraries with identical build files

custom_tvm_ops/tests/layer_norm_test.py
```diff
@@ -0,0 +1,206 @@
+import numpy
```
bwasti: this file should all live within `test/test_operators.py` or a new file `test/test_custom_operators.py`

custom_tvm_ops/topi/topi.cc
```diff
@@ -0,0 +1,47 @@
+#include <tvm/build_module.h>
```
bwasti: can all of these files live in the same directory? maybe `torch_tvm/ops/layer_norm/`?  Since this is the first operator being added I think we can keep things relatively self contained rather than adding new directory structures

torch_tvm/compiler.cpp
```diff
@@ -72,6 +72,9 @@ tvm::relay::Expr TVMCompiler::convertToRelay(
     auto v = tvm::relay::ConstantNode::make(x);
     return v;
   }
+  // TODO: This can be just NDArray of num elements of IntList.
```
bwasti: I believe that is true.  Curious, are you planning to use this fact for this PR? If not, maybe file an issue for improvement

torch_tvm/compiler.cpp
```diff
@@ -176,13 +179,16 @@ tvm::relay::Function TVMCompiler::convertToRelay(
...
 
   tvm::Array<tvm::relay::Var> free_vars = tvm::relay::FreeVars(output);
-  AT_CHECK(
-      free_vars.size() <= input_vars.size(),
-      "Determined ",
-      free_vars.size(),
-      " free vars but only ",
-      input_vars.size(),
-      " inputs");
+  // Changing this to warning because layer norm may not make use of optional weight
```
bwasti: if the number of inputs is less than the number of free variables something is broken.  Any conditional on the use of an optional weight is either 1) constant or 2) dynamic.  If it's constant we need to propagate that information and reduce the number of free variables.  Otherwise we should create a dummy variable as an input

torch_tvm/operators.cpp
```diff
@@ -269,6 +317,9 @@ RegisterTVMOperator reg({
        auto out =
            tvm::relay::CallNode::make(op, bn_inputs, tvm::Attrs(attrs), {});
        TORCH_INTERNAL_ASSERT(node->outputs().size() == 1);
+       // Is this right? Seems strange to use tupleget node if we are returning before that.
```
bwasti: I'll need to double check this -- can you file it as an issue?

custom_tvm_ops/tests/layer_norm_test.py
bwasti: agreed we certainly don't want this kind of logic if we can avoid it

kimishpatel:
@bwasti, so I only see a couple of lines of code commented out. I will remove that. But if you see more, please do point. I might have missed.

custom_tvm_ops/tests/layer_norm_test.py
kimishpatel: So I did try to fix this but since I wanted to make some forward progress I punted it for future. The other fix that include this in torch_tvm was not something I wanted to go for, about which I responded in another comment to @bwasti.

CMakeLists.txt
kimishpatel: So I do not want the custom topi ops to be specific to torch_tvm. They can be used elsewhere as well if needed, so that was the main reason to keep the structure this way. It did introduce some issues but I think it is better this way. If you think otherwise let me know and we can resolve this.

custom_tvm_ops/tests/layer_norm_test.py
kimishpatel: So this goes in the same spirit as that of the other comment. My point was to keep custom tvm ops implemented in a separate place and tested independently. They do not have any dependency on torch_tvm.

custom_tvm_ops/topi/topi.cc
kimishpatel: I followed mostly the structure from tvm, like here: https://github.com/dmlc/tvm/tree/master/topi/include/topi.

torch_tvm/compiler.cpp
kimishpatel: Nope. Not in this PR. Will file an issue for improvement.

torch_tvm/compiler.cpp
kimishpatel: @bwasti, so this was one thing I was vary of. I do want to preserve this however the way op is that is it has optional weight and bias and there is single relay op. I did not want to bloat code and add a separate relay op with weight and bias. I did see that in batch_norm this is handled differently, but I wanted to avoid adding another CallNode which I was not sure would be eliminated. If you have a suggestions I am open to it.

torch_tvm/operators.cpp
kimishpatel: @bwasti, sure.

kimishpatel:
@bwasti, I guess you probably meant not the commented code but just the comments. Yes, I plant to remove it before landing.

CMakeLists.txt
bwasti: where else are these operators used?  should we just split it out into a separate repo then?

kimishpatel: Hmm. Yeah am not sure of that. Do you prefer to move this inside torch_tvm then and maybe move it out at a later point if we decide to?

custom_tvm_ops/topi/topi.cc
bwasti: I don't think we need to follow the structure of tvm for the addition of a single operator

kimishpatel:
@bwasti, addressed most of the comments.

CMakeLists.txt
```diff
@@ -32,6 +34,7 @@ target_include_directories(_torch_tvm PUBLIC
     ${CMAKE_CURRENT_SOURCE_DIR}
     ${CMAKE_CURRENT_SOURCE_DIR}/torch_tvm
     ${TVM_DIR}/include
+    ${TVM_DIR}/src
```
bwasti: curious, what include needs this?

torch_tvm/compiler.cpp
```diff
@@ -197,13 +201,15 @@ tvm::relay::Function TVMCompiler::convertToRelay(
...
+  // bias Vars. TODO: Figure out a better way of error checking.
   TORCH_CHECK(
-      free_vars.size() <= input_vars.size(),
-      "Determined ",
-      free_vars.size(),
-      " free vars but only ",
-      input_vars.size(),
-      " inputs");
+    free_vars.size() <= input_vars.size(),
```
bwasti: can we revert this change exactly so it doesn't show on the diff?

torch_tvm/compiler.h
```diff
@@ -8,8 +8,12 @@
 #include <tvm/build_module.h>
 #include <tvm/operation.h>
 
+#include <torch/csrc/jit/ir.h>
+
 #include <vector>
 
+using namespace torch::jit;
```
bwasti: we shouldn't use `using` in header files (just add it to the cpp files)

torch_tvm/compiler.h
```diff
@@ -8,8 +8,12 @@
 #include <tvm/build_module.h>
 #include <tvm/operation.h>
 
+#include <torch/csrc/jit/ir.h>
```
bwasti: is this necessary?

torch_tvm/custom_tvm_ops/relay/custom_layer_norm.cc
```diff
@@ -0,0 +1,89 @@
...
+
+#include <cstdint>
+
+namespace tvm {
+namespace relay {
+
+Expr MakeCustomLayerNorm(Expr data, Expr gamma, Expr beta,
+    const int num_axis_to_normalize,
+    const bool affine, const double eps) {
+  //CHECK(data.as<Tensor>()->shape.size() > normalize_axis.size());
```
bwasti: rm

torch_tvm/custom_tvm_ops/tests/layer_norm_test.py
```diff
@@ -0,0 +1,200 @@
...
+        print(tvm.lower(schedule, [input_ph, output_ph], simple_mode=True))
+
+    @staticmethod
+    def optimize_schedule(schedule, output_tensor):
+        mean_var_sum = output_tensor.op.input_tensors[0]
+        divide_1 = output_tensor.op.input_tensors[1]
+        divide_2 = output_tensor.op.input_tensors[2]
+        schedule[divide_1].compute_inline()
+        schedule[divide_2].compute_inline()
+        #schedule[mean_var_sum].compute_at(s[normalized_output_ph], normalized_output_ph.op.axis[0])
```
bwasti: rm

torch_tvm/custom_tvm_ops/tests/layer_norm_test.py
```diff
@@ -0,0 +1,200 @@
...
+        weights_ph = tvm.placeholder(weight_shape, name='weights_placeholder')
+        bias_ph = tvm.placeholder(weight_shape, name='bias_placeholder')
+        affine = False
+        layer_norm = tvm.get_global_func("nn.custom_layer_norm")
+        if weight is not None and bias is not None:
+            affine = True
+        normalized_output_ph = layer_norm(input_ph, weights_ph, bias_ph, \
+                num_axis_to_normalize, affine, CustomLayerNormUtils.EPSILON_FLOAT)
+        s = tvm.create_schedule([normalized_output_ph.op])
+        #optimize_schedule(s, normalized_output_ph)
```
bwasti: rm

torch_tvm/custom_tvm_ops/topi/topi_init.cc
```diff
@@ -0,0 +1,62 @@
...
+#include "custom_layer_norm_generic_sched.h"
+#include "custom_layer_norm.h"
+
+#include <mutex>
+
+namespace tvm {
+
+using tvm::relay::OpPatternKind;
+
+void registerCustomTopiSchedules() {
```
bwasti: can you use static init instead of std::once?

torch_tvm/operators.cpp
```diff
@@ -215,6 +219,55 @@ RegisterTVMOperator reg({
...
+       for (auto& dim : normalized_axis_shape) {
+         shape.push_back(dim);
+       }
+       auto attrs = tvm::make_node<tvm::relay::CustomLayerNormAttrs>();
+       attrs->num_axis_to_normalize = normalized_axis_shape.size();
+       auto eps = static_cast<double>(relayToConstant<float>(inputs[4]));
+       attrs->eps = eps;
+
+       TVMContext ctx_;
+       ctx_.device_type = kDLCPU;
```
bwasti: hmm, how will this fair in situations where we don't have CPU guaranteed?

CMakeLists.txt
kimishpatel: This ones: https://github.com/pytorch/tvm/pull/73/files#diff-2941f7ba675a837c849ce41ea7e78ce7R4-R5

torch_tvm/custom_tvm_ops/topi/topi_init.cc
kimishpatel: @bwasti, To be able encapsulate all this in a function is kind of nice. And if it is wrapped in a function I would think call_once is a good way to ensure single initialization. I am open to suggestions.

torch_tvm/operators.cpp
kimishpatel: Do you mean what happens when CPU is not available? Or the Tensors from pytorch are not on CPU? I think that requires handling outside of all this. From the implementations of other operators it does not seem to me that we handle cases for non-CPU devices.
In fact I would think that we would need to check that the graph being lowered has inputs on CPU until we expand the scope.

bwasti: if we set the `ctx_` in the module to GPU would this still work?

torch_tvm/custom_tvm_ops/topi/topi_init.cc
bwasti: Sorry, I wasn't clear - you can use a static initialization approach rather than calling the function repeatedly and guarding the execution with std::once.  The idea boils down to defining a static variable whose constructor does the registration.

You can find this paradigm littered throughout PyTorch, and might even be able to use some of the helpers https://bddppq.github.io/codebrowser/pytorch/pytorch/c10/util/Registry.h.html#5
https://bddppq.github.io/codebrowser/pytorch/pytorch/torch/csrc/jit/custom_operator.h.html#10
https://bddppq.github.io/codebrowser/pytorch/pytorch/torch/csrc/jit/pass_manager.h.html#9

torch_tvm/operators.cpp
kimishpatel: I think it should work ok. But I wouldn't trust without testing.

torch_tvm/custom_tvm_ops/topi/topi_init.cc
kimishpatel: So I have seen these in PyTorch codebase, and I think `auto reg_ptr = tvm::runtime::Registry::Get("relay.op._Register");` does something similar. So in effect this is doing the same thing, except it ensure we dont call this method more than once. 
If you want to remove call_once, it would just mean that we remove all the function boundaries so that code is executed when the lib is loaded. We can discuss this further if you think there is a more elegant approach, but I am not convinced of it or see it yet.

torch_tvm/custom_tvm_ops/topi/topi_init.cc
```diff
@@ -0,0 +1,62 @@
...
+#include <tvm/build_module.h>
+#include <tvm/relay/attrs/transform.h>
+#include <tvm/relay/op_attr_types.h>
+#include <tvm/relay/expr.h>
+#include <tvm/runtime/packed_func.h>
+
+#include <custom_tvm_ops/relay/layer_norm_attrs.h>
+
+#include "topi_init.h"
+#include "topi/generic/injective.h"
```
llyfacebook: Add this file?

kimishpatel: @llyfacebook, didn't follow your comment. Can you elaborate?

torch_tvm/compiler.cpp
bwasti: rm this comment?

torch_tvm/operators.cpp
```diff
@@ -215,6 +219,67 @@ RegisterTVMOperator reg({
...
+       ctx_.device_id = 0;
+
+       auto val = node->input(0);
+       auto pt_t = val->type()->cast<CompleteTensorType>();
```
bwasti: `inputs[0]` has all this information, right? we should avoid using `Node*` if we can.  This is replicating the logic in `compiler.cpp` and introducing a new API (`scalarTypeToTVMType`)

torch_tvm/operators.cpp
```diff
@@ -1,3 +1,5 @@
+#include <algorithm>
```
bwasti: what's this used for?

torch_tvm/compiler.h
```diff
@@ -10,6 +10,8 @@
 
 #include <vector>
 
+tvm::relay::DataType scalarTypeToTVMType(at::ScalarType pt_type);
```
bwasti: I don't think we want this public -- is there a good reason?

torch_tvm/custom_tvm_ops/topi/topi_init.cc
bwasti: let's discuss offline

torch_tvm/operators.cpp
kimishpatel: @bwasti, from what I understand input[0] which is Value*, does not have the scalarType of tensor in it. That is why it has TypePtr. I am open to suggestions but I dont see why exposing scalarTypeToTVMType is an issue.

torch_tvm/operators.cpp
kimishpatel: good catch. Will remove it.

torch_tvm/compiler.h
kimishpatel: This is needed in operator.cpp and I think we would need it to extract TVM type in other places where right now it is hardcoded to float.

torch_tvm/custom_tvm_ops/topi/topi_init.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+namespace tvm {
+void registerCustomTopiSchedules();
```
yinghai: For OSS, python oriented user this might be a problem.

bwasti: is there a pythonic way of registering relay compute?

yinghai: Yeah, but what I'm thinking is why not upstreaming this to TVM directly?

torch_tvm/custom_tvm_ops/topi/topi_init.cc
yinghai: I agree with @bwasti with the static trick.

torch_tvm/custom_tvm_ops/topi/topi_init.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
```
bwasti: delete this file

torch_tvm/custom_tvm_ops/topi/topi_init.cc
```diff
@@ -0,0 +1,94 @@
+#include <topi/x86/default.h>
```
bwasti: this should just be `topi/custom_layer_norm.cc`

torch_tvm/custom_tvm_ops/topi/custom_layer_norm_generic_sched.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+namespace topi {
+using namespace tvm;
```
bwasti: don't use `using` in a header https://stackoverflow.com/questions/4872373/why-is-including-using-namespace-into-a-header-file-a-bad-idea-in-c

torch_tvm/custom_tvm_ops/topi/topi_init.cc
```diff
@@ -0,0 +1,94 @@
...
+#include <custom_tvm_ops/relay/layer_norm_attrs.h>
+
+#include "custom_layer_norm_generic_sched.h"
+#include "custom_layer_norm.h"
+
+namespace tvm {
+
+using tvm::relay::OpPatternKind;
+
+class CustomTOPIOpRegisterer {
```
bwasti: nit: name this `RegisterCustomLayerNorm` and make it a struct, removing the `public` just below

torch_tvm/custom_tvm_ops/topi/topi_init.h
bwasti: I'm not sure about that -- @kimishpatel ?

kimishpatel:
> torch_tvm/ops/topi/custom_layer_norm.cc # Implements and registers all topi stuff

I disagree with "torch_tvm/ops/topi/custom_layer_norm.cc # Implements and registers all topi stuff". I think we should keep implementation and registration different.

torch_tvm/custom_tvm_ops/topi/topi_init.h
kimishpatel: I thought I deleted it.

torch_tvm/custom_tvm_ops/topi/custom_layer_norm_generic_sched.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
```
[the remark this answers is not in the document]
kimishpatel: I would like to keep sched, compute and registration separate.

torch_tvm/custom_tvm_ops/tests/layer_norm_test.py
```diff
@@ -0,0 +1,199 @@
+import numpy
```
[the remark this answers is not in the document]
kimishpatel: Thats why I did not name it so. Earlier I wanted it to be a separate test that is not tied to torch_tvm tests. But I can change it if you want.

torch_tvm/custom_tvm_ops/topi/topi_init.cc
kimishpatel: Got it.

torch_tvm/custom_tvm_ops/topi/custom_layer_norm_generic_sched.h
kimishpatel: Got it.

torch_tvm/custom_tvm_ops/topi/topi_init.cc
kimishpatel: So this becomes customer layer norm specific then. I think Lingyi is going to add his registration logic corresponding to his stuff here as well, in that case `RegisterCustomLayerNorm` name would not make sense.

bwasti: that's the point of static initialization -- we avoid blockers by keeping the registration in separate compilation units rather than putting it all into a single file

torch_tvm/custom_tvm_ops/topi/custom_layer_norm_generic_sched.h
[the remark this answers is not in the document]
bwasti: that's fine, but could we move all this heavy logic away from a header

torch_tvm/custom_tvm_ops/tests/layer_norm_test.py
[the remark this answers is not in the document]
bwasti: why don't you want this to work with unittest?

torch_tvm/custom_tvm_ops/topi/topi_init.h
kimishpatel: I am ok with upstreaming it. We can discuss this.

torch_tvm/custom_tvm_ops/topi/topi_init.cc
kimishpatel: I see. I think thats fair. But note that this would also mean each op we add has its own initialization file, which I think is ok but something to keep in mind. I will rename this to custom_layer_norm_init.cc then.

torch_tvm/custom_tvm_ops/topi/custom_layer_norm_generic_sched.h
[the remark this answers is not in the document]
kimishpatel: Sure.

bwasti:
> > torch_tvm/ops/topi/custom_layer_norm.cc # Implements and registers all topi stuff
> 
> I disagree with "torch_tvm/ops/topi/custom_layer_norm.cc # Implements and registers all topi stuff". I think we should keep implementation and registration different.

that's fine, but for consistency's sake we need to split up the relay stuff too right?

torch_tvm/custom_tvm_ops/tests/layer_norm_test.py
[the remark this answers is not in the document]
kimishpatel: Actually no reason. Let me rename it.

kimishpatel:
> Assuming you are still planning to rename `topi_init.cc`?

Oh yes. Will do that before merging.

PR #75, opened 2019-07-26
tensor type node to respect pytorch dtype

wanchaol:
* **#75 tensor type node to respect pytorch dtype**


torch_tvm/compiler.cpp
```diff
@@ -8,6 +8,23 @@
...
 
+
+tvm::relay::DataType scalarTypeToTVMType(at::ScalarType pt_type) {
+  static const std::unordered_map<at::ScalarType, tvm::relay::DataType> type_mapping = {
+    {at::ScalarType::Double, ::tvm::Float(64)},
+    {at::ScalarType::Float, ::tvm::Float(32)},
+    {at::ScalarType::Int, ::tvm::Int(32)},
+    {at::ScalarType::Long, ::tvm::Int(64)},
+    {at::ScalarType::Bool, ::tvm::Bool()},
+    {at::ScalarType::Byte, ::tvm::UInt(8)},
```
llyfacebook: Can you also map the Qtensor to int8? By the way, how to differentiate the uint8 and int8 in ScalarType?

wanchaol: scalartype::char is int8 while byte is uint8 according to https://github.com/pytorch/pytorch/blob/master/c10/core/ScalarType.h

lly-zero-one (approved):

PR #82, opened 2019-08-07
Conv1D and reduce op max.

kimishpatel:
Added support for 1D conv.
Added mapping of torch reduction max op.

### DIFF
--- a/torch_tvm/operators.cpp
+++ b/torch_tvm/operators.cpp
@@ -1,8 +1,5 @@
-#include "operators.h"
 #include <tvm/relay/attrs/nn.h>
 #include <tvm/relay/attrs/transform.h>
-#include "compiler.h"
-#include "fusion_pass.h" // tvm_sym
 
 #include <torch/csrc/autograd/record_function.h>
 #include <torch/csrc/jit/custom_operator.h>
@@ -10,6 +7,9 @@
 #include <torch/csrc/jit/passes/utils/subgraph_utils.h>
 
 #include "custom_tvm_ops/relay/custom_layer_norm_attrs.h"
+#include "compiler.h"
+#include "fusion_pass.h" // tvm_sym
+#include "operators.h"
 
 using namespace torch::jit;
 
@@ -98,6 +98,9 @@ T relayToConstant(tvm::relay::Expr e) {
   auto c = e.as<tvm::relay::ConstantNode>();
   TORCH_INTERNAL_ASSERT(c);
   TORCH_INTERNAL_ASSERT(c->is_scalar());
+  // TODO: Better error handling on checking if T is same as
+  // c->data->dtype. Unfortunately DLDataType is enum of
+  // int, uint and float.
   return static_cast<T*>(c->data->data)[0];
 }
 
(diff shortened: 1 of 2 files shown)


torch_tvm/operators.cpp
```diff
@@ -117,6 +120,23 @@ uint64_t getNoneSentinel() {
   return 0xe4fa3adecabcf036;
 }
 
+tvm::relay::Expr insertDims(const tvm::relay::Expr& in, int64_t input_dims,
```
bwasti: may as well register these as ops?

torch_tvm/operators.cpp
```diff
@@ -183,27 +217,60 @@ RegisterTVMOperator reg({
...
        if (const tvm::relay::VarNode* var =
                inputs[1].as<tvm::relay::VarNode>()) {
          auto* w_t = var->type_annotation.as<tvm::relay::TensorTypeNode>();
          TORCH_INTERNAL_ASSERT(w_t);
          auto shape = w_t->shape;
-         tvm::Array<tvm::relay::IndexExpr> w_sizes = {shape[2], shape[3]};
+         auto num_dims = shape.size();
+         tvm::Array<tvm::relay::IndexExpr> w_sizes;
+         if (num_dims < 4) {
+           AT_CHECK(num_dims == 3,
```
bwasti: I think we need to use the new CHECK api (TORCH_CHECK?)

torch_tvm/operators.cpp
```diff
@@ -183,27 +217,60 @@ RegisterTVMOperator reg({
...
        conv_attrs->kernel_layout = "OIHW";
+       conv_attrs->strides = relayToArray<tvm::relay::IndexExpr>(inputs[3]);
+       conv_attrs->padding = relayToArray<tvm::relay::IndexExpr>(inputs[4]);
+       conv_attrs->dilation = relayToArray<tvm::relay::IndexExpr>(inputs[5]);
 
        conv_attrs->kernel_size =
            tvm::NullValue<tvm::Array<tvm::relay::IndexExpr>>();
+       // For 1D conv need to expand dim. That means it has to be removed
+       // as well.
+       bool dim_added{false}; // For 1D conv need to expand dim.
```
bwasti: this is fine for now, but I wonder if we should be using a pass in PyTorch to make this more portable for other backends going forward? maybe add a TODO

torch_tvm/operators.cpp
```diff
@@ -212,8 +279,8 @@ RegisterTVMOperator reg({
          auto bias_add_op = tvm::relay::Op::Get("nn.bias_add");
          auto bias_add_attrs = tvm::make_node<tvm::relay::BiasAddAttrs>();
          bias_add_attrs->axis = 1;
-         return tvm::relay::CallNode::make(
-             bias_add_op, {out, inputs[2]}, tvm::Attrs(bias_add_attrs), {});
+         return tvm::relay::Expr(tvm::relay::CallNode::make(
```
bwasti: we shouldn't need to wrap CallNode with Expr

torch_tvm/operators.cpp
```diff
@@ -183,27 +217,60 @@ RegisterTVMOperator reg({
        conv_attrs->groups = relayToConstant<int>(inputs[8]);
        conv_attrs->data_layout = "NCHW";
        conv_attrs->kernel_layout = "OIHW";
```
llyfacebook: This seems to be invalid for 1d conv, right?

kimishpatel: @llyfacebook, we are converting it to 2D conv so it is still valid.

llyfacebook: ok, I assume you could still distinguish it from conv2d in tvm side, since you will need to specialize the perf optimization for conv1d.

kimishpatel: @llyfacebook, tvm has no conv1d op. However neither does mkldnn as far as I understand. We will have to just benchmark and see how they perform.

torch_tvm/operators.cpp
kimishpatel: @bwasti, this is just a helper function to wrap to expand_dims.

torch_tvm/operators.cpp
kimishpatel: @bwasti, not sure what you mean. Do you mean make 1d conv look like 2D with jit pass and use such a pass if needed?

torch_tvm/operators.cpp
kimishpatel: There is not TORCH_CHECK and the only other one is internal_assert.

torch_tvm/operators.cpp
kimishpatel: Thats what I thought, but it did not work in the first place. I changed the signature of the wrapper functions and now it seems fine.

PR #83, opened 2019-08-08
Add the support of lowering torch.cat to relay::concatenate

llyfacebook:


test/test_operators.py
```diff
@@ -284,6 +284,24 @@ def linear_no_bias(input, weight):
         ref_out_no_bias, tvm_out_no_bias = self.runBoth(linear_no_bias, input, weight)
         assert torch.allclose(ref_out_no_bias, tvm_out_no_bias, rtol=0.01, atol=0.01)
 
+    @TVMTest.given(
+        shape=TVMTest.rand_shape(rank=2, min_dim=4),
+    )
+    def test_concat(self, shape):
+        input1 = torch.rand(shape)
+        input2 = torch.rand(shape)
+        # import pdb
```
kimishpatel: Remove this.

test/test_operators.py
```diff
@@ -284,6 +284,24 @@ def linear_no_bias(input, weight):
...
+        # pdb.set_trace()
+        def concat(x1, x2):
+            return torch.cat((x1, x2), 0)
+
+        import torch_tvm
+        torch_tvm.enable()
+        # test concat
+        scripted_concat = torch.jit.script(concat)
+        concat_graph = scripted_concat.graph_for(input1, input2)
+        FileCheck().check("prim::FusedConcat").check_not("prim::ListConstruct").check_not("aten::cat").run(str(concat_graph))
```
kimishpatel: This seems to be the test not for concat but whether ListConstruct and cat are fuse. We should name it such.

test/test_operators.py
```diff
@@ -284,6 +284,24 @@ def linear_no_bias(input, weight):
...
+        input2 = torch.rand(shape)
+        # import pdb
+        # pdb.set_trace()
+        def concat(x1, x2):
+            return torch.cat((x1, x2), 0)
+
+        import torch_tvm
+        torch_tvm.enable()
+        # test concat
+        scripted_concat = torch.jit.script(concat)
```
kimishpatel: Also add a test for actual concat? You can use the same runBoth functionality for that.

torch_tvm/fuse_concat.cpp
```diff
@@ -0,0 +1,59 @@
...
+using namespace torch::jit;
+
+bool isFusableCatNode(Node* node) {
+  if (node->kind() != aten::cat)
+    return false;
+  if (!node->is_constant(attr::dim))
+    return false;
+
+  auto tensors_node = node->namedInput(attr::tensors)->node();
+  if ((tensors_node->inputs().size() + node->outputs().size()) >
```
kimishpatel: Can you add a comment as to what is this supposed to check? And maybe `constexpr int some_var_name = 128`?

torch_tvm/fuse_concat.cpp
```diff
@@ -0,0 +1,59 @@
...
+  AT_ASSERT(node->kind() == aten::cat);
+  Graph* graph = node->owningGraph();
+  Node* list_construct = node->namedInput(attr::tensors)->node();
+  int64_t dim = node->get<int64_t>(attr::dim).value();
+
+  Node* fused_cat = graph->create(prim::FusedConcat, list_construct->inputs())
+                        ->i_(attr::dim, dim);
+  fused_cat->insertBefore(list_construct);
+  fused_cat->output()->copyMetadata(node->output());
+  node->output()->replaceAllUsesWith(fused_cat->output());
```
kimishpatel: Dont you want to check `node->output()->uses().empty()` and erase that as well?

torch_tvm/fuse_concat.cpp
```diff
@@ -0,0 +1,59 @@
...
+  if (list_construct->output()->uses().empty()) {
```
kimishpatel: Don't you have to first check this condition on node, destroy that and check it again on list_construct? Or does pytorch remove all the nodes without uses on its own?

torch_tvm/fuse_concat.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+void fuseConcats(Block* block_, AliasDb& aliasDb) {
+  for (auto it = block_->nodes().rbegin(); it != block_->nodes().rend();
+       ++it) {
+    Node* cat = *it;
+    if (!isFusableCatNode(cat)) {
+      continue;
+    }
+
+    Node* list_construct = cat->namedInput(attr::tensors)->node();
```
kimishpatel: Does not seem like `list_construct` is being used.

torch_tvm/fuse_concat.cpp
```diff
@@ -0,0 +1,59 @@
...
+    createFusedConcat(cat);
+  }
+}
+
+void FuseConcat(std::shared_ptr<Graph> graph) {
```
kimishpatel: @bwasti, should we construct aliasdb once and run all the fusion passes?

torch_tvm/fuse_concat.cpp
```diff
@@ -0,0 +1,59 @@
...
+  fused_cat->insertBefore(list_construct);
+  fused_cat->output()->copyMetadata(node->output());
+  node->output()->replaceAllUsesWith(fused_cat->output());
+  if (list_construct->output()->uses().empty()) {
+    list_construct->destroy();
+  }
+  return fused_cat;
+}
+
+void fuseConcats(Block* block_, AliasDb& aliasDb) {
```
kimishpatel: It doesnt seem that aliasDb is being used. Remove it.

torch_tvm/fuse_concat.cpp
```diff
@@ -0,0 +1,59 @@
...
+      continue;
+    }
+
+    Node* list_construct = cat->namedInput(attr::tensors)->node();
+    createFusedConcat(cat);
+  }
+}
+
+void FuseConcat(std::shared_ptr<Graph> graph) {
+  AliasDb aliasDb(graph);
```
kimishpatel: Since it is not used this can be removed too.

kimishpatel: @bwasti, should we add unused-variable as error flag?

torch_tvm/fuse_concat.cpp
yinghai: This is fine. It's done once.

torch_tvm/fuse_concat.cpp
```diff
@@ -0,0 +1,59 @@
+#include "fusion_pass.h"
+#include "operators.h"
+
+using namespace torch::jit;
+
+bool isFusableCatNode(Node* node) {
+  if (node->kind() != aten::cat)
```
yinghai: Please add `{}` around your block.

torch_tvm/fuse_concat.cpp
kimishpatel: I also suppose, once the graph is altered by earlier fusions aliasing information will change. So makes sense to do it for each fusion pass.

yinghai:
@llyfacebook Let's land this?

test/test_operators.py
llyfacebook: sure

torch_tvm/fuse_concat.cpp
llyfacebook: I extracted some code from pytorch side and didn't clean it up. :)

torch_tvm/fuse_concat.cpp
llyfacebook: I think the replaceAllUsesWith do it internally.

torch_tvm/fuse_concat.cpp
```diff
@@ -0,0 +1,65 @@
...
+    list_construct->destroy();
+  }
+  return fused_cat;
+}
+
+void fuseConcats(Block* block_) {
+  for (auto it = block_->nodes().rbegin(); it != block_->nodes().rend();
+       ++it) {
+    Node* cat = *it;
+    if (!isFusableCatNode(cat)) {
```
kimishpatel: nit. Can be just `!isFusableCatNode(*it)`.

PR #84, opened 2019-08-12
Add the quantization linear op support

llyfacebook:


torch_tvm/custom_tvm_ops/topi/contrib/quantize.cc
```diff
@@ -0,0 +1,78 @@
...
+
+using namespace runtime;
+
+TVM_REGISTER_GLOBAL("tvm.contrib.find_minmax")
+    .set_body([](TVMArgs args, TVMRetValue* ret) {
+      DLTensor* input = args[0];
+      DLTensor* data_min = args[1];
+      DLTensor* data_max = args[2];
+      // calculate the data_min and data_max
+      auto data_ptr = static_cast<float *>(input->data);
```
bwasti: DLTensor has strides, should we assert those aren't used before doing this contiguous sweep?

torch_tvm/custom_tvm_ops/topi/custom_topi_ops.cc
```diff
@@ -0,0 +1,248 @@
+#include <topi/generic/injective.h>
```
bwasti: why are we merging everything into one file? this will probably make rebasing hard going forward.  you could just copy paste the custom_layer_norm_init.cc file and replace it with your stuff

CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ file(GLOB TORCH_TVM_SRCS
   ${CMAKE_CURRENT_SOURCE_DIR}/torch_tvm/*.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/torch_tvm/custom_tvm_ops/relay/*.cc
   ${CMAKE_CURRENT_SOURCE_DIR}/torch_tvm/custom_tvm_ops/topi/*.cc
+  ${CMAKE_CURRENT_SOURCE_DIR}/torch_tvm/custom_tvm_ops/topi/contrib/*.cc
```
bwasti: probably fine to leave it in topi/* without adding a contrib folder

torch_tvm/custom_tvm_ops/topi/quantize.cc
```diff
@@ -0,0 +1,103 @@
...
+          return scale_mul(0)*(tvm::cast(Int(32), quantized_mm(i, j)) -
+              tvm::cast(Float(32), data_acc(i))*weight_zero_point(0) -
+              tvm::cast(Float(32), weight_acc(i))*data_zero_point(0) +
+              zero_point_mul(0));
+        },
+        "tensor",
+        "mm_dequantize"
+      );
+  //TODO: add Bias
+  return {result};
```
yinghai: Maybe you don't need all the tvm::compute above and return them as a tensor here directly.

CMakeLists.txt
llyfacebook: It seems not working.

torch_tvm/custom_tvm_ops/topi/custom_topi_ops.cc
llyfacebook: Hmm, I actually prefer a centralized place to store all these registration. @yinghai I think c2 does the same thing, right?

yinghai: Yeah, just a place to see how many ops we have.

torch_tvm/custom_tvm_ops/relay/quantize.cc
```diff
@@ -0,0 +1,148 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/*!
+ *  Copyright (c) 2018 by Contributors
+ * \file nn.cc
+ * \brief Property def of nn operators.
+ */
+
```
kimishpatel: This is not licensed under anything, is it?

test/test_operators.py
```diff
@@ -326,6 +326,27 @@ def linear_no_bias(input, weight):
...
+        bias = torch.rand(out_features)
+        q_weight, col_offsets, scale, zero_point = \
+            torch.fbgemm_linear_quantize_weight(weight.clone().float())
+        packed_weight = torch.fbgemm_pack_quantized_matrix(q_weight.clone())
+
+        def fbgemm_quantized_linear(input, weight, bias):
+            return torch.fbgemm_linear_int8_weight_fp32_activation(
+                input.float(), q_weight, packed_weight, col_offsets, scale, zero_point, bias.float())
+        ref_out, tvm_out = self.runBoth(fbgemm_quantized_linear, input, weight, bias)
+        # relax the constraint to abvoid flaky test
```
kimishpatel: nit. "abvoid" -> avoid.

torch_tvm/custom_tvm_ops/relay/quantize.cc
```diff
@@ -0,0 +1,148 @@
...
+#include <relay/op/op_common.h>
+#include <relay/op/type_relations.h>
+#include <relay/pass/alter_op_layout.h>
+#include "quantize_attrs.h"
+
+#include <cstdint>
+namespace tvm {
+namespace relay {
+
+Expr MakeDataInt8Quantization(Expr data, Expr zero_point, Expr scale, bool is_signed, int precision) {
```
kimishpatel: What is the use of precision here? It seems Int8 is already telling? Or is it for output precision?

torch_tvm/custom_tvm_ops/relay/quantize.cc
```diff
@@ -0,0 +1,148 @@
...
+  if(param->is_signed) {
+    out_dtype = Int(param->precision);
+  } else {
+    out_dtype = UInt(param->precision);
+  }
+  std::vector<Type> fields;
+  fields.push_back(TensorTypeNode::make(oshape, out_dtype));
+  fields.push_back(TensorTypeNode::make(acc_oshape, Int(32)));
+  reporter->Assign(types[3], TupleTypeNode::make(Array<Type>(fields)));
+  //  reporter->Assign(types[3], TensorTypeNode::make(oshape, out_dtype));
```
kimishpatel: rm

torch_tvm/custom_tvm_ops/relay/quantize.cc
llyfacebook: hmm, just in case, for 16bits, for example.

torch_tvm/custom_tvm_ops/relay/quantize.h
```diff
@@ -0,0 +1,61 @@
+#include <relay/op/op_common.h>
```
kimishpatel: `#pragma once`

llyfacebook: okok

torch_tvm/custom_tvm_ops/relay/quantize.cc
```diff
@@ -0,0 +1,148 @@
...
+#include <relay/pass/alter_op_layout.h>
+#include "quantize_attrs.h"
+
+#include <cstdint>
+namespace tvm {
+namespace relay {
+
+Expr MakeDataInt8Quantization(Expr data, Expr zero_point, Expr scale, bool is_signed, int precision) {
+  static const Op& op = Op::Get("nn.quantize_data_int8_quantize");
+  auto attrs = make_node<QuantizedParamsAttrs>();
```
kimishpatel: I am little queazy about QuantizedParamsAttrs also storing zp and scale. Here it seems to be used for generating zero point and scale, but it can also store weight's zp and scale. I am not insistent on it, but should we split it into `DataRepAttrs` (data representation attrs) and `QuantizatioAttrs`? What I am suggesting is also not clean, but something to think over. Either way I am not too picky on this.

torch_tvm/custom_tvm_ops/relay/quantize_init.cc
```diff
@@ -0,0 +1,82 @@
...
+
+TVM_REGISTER_NODE_TYPE(QuantizedParamsAttrs);
+
+TVM_REGISTER_API("relay.op.nn._make.quantize_data_int8_quantize")
+  .set_body_typed(MakeDataInt8Quantization);
+
+
+RELAY_REGISTER_OP("nn.quantize_data_int8_quantize")
+.describe(R"code(dynamic quantization of weight or activation.
+- **weight**: (channels, in_channels)
```
kimishpatel: I dont follow what this is supposed to mean.

torch_tvm/custom_tvm_ops/relay/quantize_init.cc
```diff
@@ -0,0 +1,82 @@
...
+namespace relay {
+
+TVM_REGISTER_NODE_TYPE(QuantizedParamsAttrs);
+
+TVM_REGISTER_API("relay.op.nn._make.quantize_data_int8_quantize")
+  .set_body_typed(MakeDataInt8Quantization);
+
+
+RELAY_REGISTER_OP("nn.quantize_data_int8_quantize")
+.describe(R"code(dynamic quantization of weight or activation.
```
kimishpatel: This seems to be specific for data.

torch_tvm/custom_tvm_ops/topi/contrib/quantize.cc
kimishpatel: @bwasti, for minmax it should not matter though, right? Order in which elements are traversed is not important.

torch_tvm/custom_tvm_ops/topi/contrib/quantize.cc
```diff
@@ -0,0 +1,80 @@
...
+TVM_REGISTER_GLOBAL("tvm.contrib.find_minmax")
+    .set_body([](TVMArgs args, TVMRetValue* ret) {
+      DLTensor* input = args[0];
+      DLTensor* data_min = args[1];
+      DLTensor* data_max = args[2];
+      // calculate the data_min and data_max
+      CHECK(input->strides == nullptr) << "find_minmax does not support the dltensor with strides";
+      auto data_ptr = static_cast<float *>(input->data);
+      int m = input->shape[0];
+      int n = input->shape[1];
```
kimishpatel: We need to asset that tensor is 2-dim.

torch_tvm/custom_tvm_ops/topi/contrib/quantize.cc
```diff
@@ -0,0 +1,80 @@
...
+      float d_min = data_ptr[0];
+      float d_max = data_ptr[0];
+      for (int i = 0; i < m; i++) {
+        for (int j = 0; j < n; j++) {
+            d_min = std::min(data_ptr[i*n + j], d_min);
+            d_max = std::max(data_ptr[i*n + j], d_max);
+        }
+      }
+      auto out_ptr_min = static_cast<float *>(data_min->data);
```
kimishpatel: reinterpret_cast?

torch_tvm/custom_tvm_ops/topi/custom_topi_ops.cc
```diff
@@ -0,0 +1,205 @@
...
+               const tvm::Array<tvm::Tensor>& outs,
+               const tvm::Target& target) -> tvm::Schedule {
+              // dispatches x86 schedule when the target is "llvm" and the
+              return topi::generic::schedule_custom_layer_norm(outs);
+            }),
+        10);
+    (*reg_ptr)(
+        "nn.quantize_findminmax",
+        "TOpPattern",
+        static_cast<int>(OpPatternKind::kOutEWiseFusable),
```
kimishpatel: Is findminmax elementwise fusable? I dont think so right?

torch_tvm/custom_tvm_ops/topi/custom_topi_ops.cc
```diff
@@ -0,0 +1,205 @@
...
+            [](const tvm::Attrs& attrs,
+               const tvm::Array<tvm::Tensor>& outs,
+               const tvm::Target& target) -> tvm::Schedule {
+              return  topi::generic::schedule_quantize_findminmax(outs);
+            }),
+        10);
+    (*reg_ptr)(
+        "nn.choose_quantize_params",
+        "TOpPattern",
+        static_cast<int>(OpPatternKind::kOutEWiseFusable),
```
kimishpatel: Same comment in other places. Is the op pattern right?

torch_tvm/custom_tvm_ops/topi/custom_topi_ops.cc
```diff
@@ -0,0 +1,205 @@
...
+               const tvm::relay::Type& out_type,
+               const tvm::Target& target) -> tvm::Array<tvm::Tensor> {
+              const auto* param = attrs.as<relay::QuantizedParamsAttrs>();
+              return tvm::Array<tvm::Tensor>{topi::data_int8_mm_dequantize(
+                inputs[0], inputs[1], inputs[2], inputs[3],
+                inputs[4], inputs[5], param->w_scale, param->w_zp)};
+            }),
+        10);
+    (*reg_ptr)(
+         "nn.quantize_data_mm_dequantize",
```
kimishpatel: Indentation.

torch_tvm/custom_tvm_ops/topi/quantize.cc
```diff
@@ -0,0 +1,89 @@
+#pragma once
```
kimishpatel: No `pragma once` here.

torch_tvm/custom_tvm_ops/topi/quantize.cc
```diff
@@ -0,0 +1,89 @@
...
+
+namespace topi {
+using namespace tvm;
+
+Array<Tensor> data_int8_quantize(
+    const Tensor& data,
+    const Tensor& zero_point,
+    const Tensor& scale,
+    bool is_signed,
+    int precision) {
```
kimishpatel: [minor] If this is int8, precision is known already. Should we rename the function appropriately? Or just remove precision?

torch_tvm/custom_tvm_ops/topi/custom_topi_ops.cc
llyfacebook: Yes, need to take care of these about the pattern.

torch_tvm/custom_tvm_ops/topi/quantize.cc
llyfacebook: We should allow it for fp16 in the future, for example. The name of the function needs to be generalized.

torch_tvm/custom_tvm_ops/topi/quantize.cc
llyfacebook: Good catch! I guess I copied it from .h file.

torch_tvm/operators.cpp
```diff
@@ -404,6 +405,51 @@ RegisterTVMOperator reg({
        n->index = 0;
        return tvm::relay::TupleGetItem(n);
      }},
+    {Symbol::fromQualString("aten::fbgemm_linear_int8_weight_fp32_activation"),
```
kimishpatel: This is too big. Let's make it into a function like `lowerFbgemmLinearInt8`

torch_tvm/custom_tvm_ops/topi/custom_topi_ops.cc
```diff
@@ -0,0 +1,205 @@
...
+        10);
+    (*reg_ptr)(
+        "nn.quantize_findminmax",
+        "FTVMCompute",
+        tvm::relay::FTVMCompute(
+            [](const tvm::Attrs& attrs,
+               const tvm::Array<tvm::Tensor>& inputs,
+               const tvm::relay::Type& out_type,
+               const tvm::Target& target) -> tvm::Array<tvm::Tensor> {
+              return tvm::Array<tvm::Tensor>{topi::contrib::quantize_findminmax(inputs[0])};
```
kimishpatel: No need for `tvm::Array<tvm::Tensor>{`? `topi::contrib::quantize_findminmax` already returns array of Tensors?

torch_tvm/custom_tvm_ops/topi/custom_topi_ops.cc
```diff
@@ -0,0 +1,205 @@
...
+        "FTVMCompute",
+        tvm::relay::FTVMCompute(
+            [](const tvm::Attrs& attrs,
+               const tvm::Array<tvm::Tensor>& inputs,
+               const tvm::relay::Type& out_type,
+               const tvm::Target& target) -> tvm::Array<tvm::Tensor> {
+              const auto* param = attrs.as<relay::QuantizedParamsAttrs>();
+              auto precision = param->precision;
+              auto is_signed = param->is_signed;
+              return tvm::Array<tvm::Tensor>{topi::contrib::choose_quantize_params(
```
kimishpatel: Same here.

torch_tvm/operators.cpp
```diff
@@ -404,6 +405,51 @@ RegisterTVMOperator reg({
...
+       // inputs[3]: weight_acc
+       // inputs[4]: weight_scale
+       // inputs[5]: weight_zp
+       // inputs[6]: bias
+       TORCH_CHECK(
+           inputs.size() == 7, "Given input size from quantized linear", inputs.size(), " inputs");
+       auto op_find_minmax = tvm::relay::Op::Get("nn.quantize_findminmax");
+       auto op_choose_q_params = tvm::relay::Op::Get("nn.choose_quantize_params");
+       auto op_data_q = tvm::relay::Op::Get("nn.quantize_data_int8_quantize");
+       auto op_data_deq = tvm::relay::Op::Get("nn.quantize_data_mm_dequantize");
```
kimishpatel: Should we make dequantize a separate relayop? I am wondering if we can fuse dequantize and quantize of the next layer somehow. Or figure out some way to fuse those?

torch_tvm/custom_tvm_ops/topi/contrib/quantize.cc
llyfacebook: I think it is void pointer. Static_cast should be better?

torch_tvm/operators.cpp
kimishpatel: Oh the bias is not quantized? Maybe we can try to quantize bias and weight as a joint quantity? Something to think over.

torch_tvm/custom_tvm_ops/topi/contrib/quantize.cc
llyfacebook: IIRC, it should be void pointer.

torch_tvm/operators.cpp
llyfacebook: Hmm, this involves the batchnorm issue. Currently, it is following the dynamic quantization schema.

llyfacebook: > Should we make dequantize a separate relayop? I am wondering if we can fuse dequantize and quantize of the next layer somehow. Or figure out some way to fuse those?

Normally, I think what you are saying is more like static quantization. We can collapse the quant and dequant, since they are the same q parameters. Dequant is normally fused with the op.

torch_tvm/custom_tvm_ops/topi/contrib/quantize.cc
kimishpatel: I suppose static_cast is safer

torch_tvm/operators.cpp
kimishpatel: Aah, yes. You are right. But if we are doing dequant just to find the min max values and then chose quantization params and quantize it again then can there not be a way to find fp32 min max values without going though dequant step? Dequant step should not be changing which value is min and which is max?

torch_tvm/custom_tvm_ops/topi/contrib/quantize.cc
llyfacebook: Yes

torch_tvm/operators.cpp
llyfacebook: I didn't get your point. Basically, dequant in this context is for int32 activation to fp32 conversion.

PR #85, opened 2019-08-12
Removed duplicate node creation.

kimishpatel:
Summary:
In current implementation if the node has multiple inputs then for each input we query its uses and for each use we try to create a node even though the node might have been created by uses of another input. This PR fixes that.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torch_tvm/compiler.cpp
+++ b/torch_tvm/compiler.cpp
@@ -148,6 +148,25 @@ tvm::relay::Function TVMCompiler::convertToRelay(
       auto uses = value->uses();
       for (const auto& use : uses) {
         tvm::Array<tvm::relay::Expr> relay_inputs;
+        if (use.user->outputs().size() == 1) {
+          if (value_map.count(use.user->output())) {
+            continue;
+          }
+        } else {
+          bool node_already_inserted{false};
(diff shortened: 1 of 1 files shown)


torch_tvm/compiler.cpp
```diff
@@ -148,6 +148,25 @@ tvm::relay::Function TVMCompiler::convertToRelay(
       auto uses = value->uses();
       for (const auto& use : uses) {
         tvm::Array<tvm::relay::Expr> relay_inputs;
+        if (use.user->outputs().size() == 1) {
+          if (value_map.count(use.user->output())) {
+            continue;
+          }
+        } else {
+          bool node_already_inserted{false};
+          for (const auto& output : use.user->outputs()) {
```
bwasti: what about

```
if (std::any_of(use.user->outputs().begin(), use.user->outputs().end(),
    [&value_map](const auto& output){return value_map.count(output);})) {
   continue;
}
```

kimishpatel: looks great. Will replace with this.

kimishpatel: @bwasti, made the change.

Issue #80, opened 2019-08-05
Memory misalignment during Tensor conversion


yinghai:
PyTorch uses offset to data pointers for `view`. This will break the 64-byte alignment requirement of TVM's input tensors, especially when we combine this we `set_input_zero_copy`. Temporary solution would be check the alignment of the data point and make an alignment copy if it's not aligned.

PR #86, opened 2019-08-12
Managed aligned memory

kimishpatel:
1. Managed DLTensor when input PT Tensor is unaligned.
       It maintains allocation and eallocation DLTensors corresponding
       PT Values and maintains a map of it.
    2. Managed param tensor just wraps that and assumes that parameters
       are immutable. Thus if a param tensor is created using
       ManagedParamTensors then it will return a cached copy of the tensor
       corresponding to the Value* if one exists. If not allocation and copy
       from corresponding PT tensor is performed. This happens only if
       input PT tensor is unaligned.
    3. Adds a way to annotate ops to specify which of their inputs are
       parameters that are expected to be immutable.

Need to add a few test to validate no memory leaks.

### DIFF
--- a/torch_tvm/compiler.cpp
+++ b/torch_tvm/compiler.cpp
@@ -5,8 +5,11 @@
 #include <torch/csrc/jit/constants.h>
 #include <torch/csrc/jit/interpreter.h>
 #include <limits>
+#include <tvm/runtime/device_api.h>
 
 using namespace torch::jit;
+
+using torch_tvm::utils::DLManagedTensorPtr;
 
 tvm::relay::DataType scalarTypeToTVMType(at::ScalarType pt_type) {
   static const std::unordered_map<at::ScalarType, tvm::relay::DataType> type_mapping = {
@@ -138,7 +141,7 @@ tvm::relay::Expr TVMCompiler::convertToRelay(
 tvm::relay::Function TVMCompiler::convertToRelay(
     std::shared_ptr<Graph> subgraph,
     TVMContext ctx,
-    std::vector<Value*>* input_values) {
+    std::vector<std::pair<Value*, bool>>* input_values) {
   std::unordered_map<Value*, tvm::relay::Expr> value_map;
   tvm::Array<tvm::relay::Var> input_vars;
 
@@ -147,7 +150,8 @@ tvm::relay::Function TVMCompiler::convertToRelay(
     auto v = convertToRelay(input, ctx);
     input_vars.push_back(v);
     if (input_values) {
-      input_values->emplace_back(input);
+      // Primary inputs are always mutable.
+      input_values->emplace_back(input, false);
     }
     value_map[input] = v;
   }
@@ -160,11 +164,19 @@ tvm::relay::Function TVMCompiler::convertToRelay(
       auto uses = value->uses();
       for (const auto& use : uses) {
         tvm::Array<tvm::relay::Expr> relay_inputs;
+        // Things like prim::Return
+        // Should we be more explicit here?
+        // That only prim::Return should be skipped?
+        if (use.user->outputs().size() < 1) {
+          continue;
+        }
         auto skip_user = false;
         if (std::any_of(use.user->outputs().begin(), use.user->outputs().end(),
               [&value_map](Value* const output){return value_map.count(output);})) {
           continue;
         }
+        const auto& param_indices = getParamIndices(use.user);
+        int input_index{0};
         for (const auto& input : use.user->inputs()) {
           if (value_map.find(input) == value_map.end()) {
             // We may be dealing with a constant, handle that here
@@ -175,7 +187,7 @@ tvm::relay::Function TVMCompiler::convertToRelay(
             } else {
               if (optional_ivalue.value().isTensor()) {
                 if (input_values) {
-                  input_values->emplace_back(input);
+                  input_values->emplace_back(input, false);
                 }
                 auto input_var = convertToRelay(input, ctx);
                 input_vars.push_back(input_var);
@@ -185,13 +197,25 @@ tvm::relay::Function TVMCompiler::convertToRelay(
               }
             }
           }
+          // Annotate the value: Whether the Value corresponds to parameter
+          // and thus is expected to be immutable.
+          if (!skip_user && input_values &&
+              std::find(param_indices.begin(),
+                param_indices.end(), input_index) != param_indices.end()) {
+            auto it = std::find_if(input_values->begin(), input_values->end(),
+                [input](const std::pair<Value*, bool>& v)
+                {return v.first == input;});
+            // It is possible that parameter is not in input_values.
+            // This can happen for example when bias input is None in
+            // convolution.
+            if (it != input_values->end()) {
+              (*it).second = true;
+            }
+          }
           relay_inputs.push_back(value_map[input]);
+          input_index++;
         }
         if (skip_user) {
-          continue;
-        }
-        // Things like prim::Return
-        if (use.user->outputs().size() < 1) {
           continue;
         }
         // if there are 2+ outputs, getOperator returns a tuple
@@ -328,16 +352,42 @@ void TVMCompiler::run(Stack& stack) {
         "Compiled subgraph with mismatching num outputs");
   }
 
+  // Using vector of unique pointers with custom deleter to
+  // delete allocated memory when gone out of scope.
+  // Only for those inputs which are not parameters.
+  // Parameters are managed by cached tvm_param_tensors.
+  // They get deallocated when cache_ is deleted.
+  std::vector<DLManagedTensorPtr> dl_tensor_list;
   for (auto i = 0; i < cache_[spec].input_values.size(); ++i) {
-    auto* value = cache_[spec].input_values[i];
+    auto* value = cache_[spec].input_values[i].first;
+    bool input_immutable = cache_[spec].input_values[i].second;
     if (!value_to_ivalue.count(value)) {
       auto optional_ivalue = toIValue(value);
       TORCH_INTERNAL_ASSERT(optional_ivalue.has_value());
       value_to_ivalue[value] = optional_ivalue.value();
     }
-    auto ivalue = value_to_ivalue.at(cache_[spec].input_values[i]);
-    auto tensor = ivalue.toTensor().to(at::kFloat);
-    auto dl_tensor = at::toDLPack(tensor);
+    auto ivalue = value_to_ivalue.at(cache_[spec].input_values[i].first);
+    //auto tensor = ivalue.toTensor().to(at::kFloat);
+    auto tensor = ivalue.toTensor();
+    DLManagedTensor* dl_tensor;
+    if (tensor.is_contiguous() &&
+        torch_tvm::utils::isAligned(tensor.data_ptr(),
+          tvm::runtime::kAllocAlignment)) {
+      dl_tensor = at::toDLPack(tensor);
+    } else if (input_immutable) {
+      if (cache_[spec].tvm_param_tensors.find(value) ==
+        cache_[spec].tvm_param_tensors.end()) {
+        dl_tensor = torch_tvm::utils::allocAndCopyData(tensor);
+        cache_[spec].tvm_param_tensors[value] = DLManagedTensorPtr(dl_tensor);
+      } else {
+        dl_tensor = cache_[spec].tvm_param_tensors[value].get();
+      }
+    } else {
+      dl_tensor =
+        torch_tvm::utils::allocAndCopyData(tensor);
+      dl_tensor_list.emplace_back(
+          dl_tensor);
+    }
     cache_[spec].set_input(i, tvm::runtime::NDArray::FromDLPack(dl_tensor));
   }
 
--- a/torch_tvm/compiler.h
+++ b/torch_tvm/compiler.h
@@ -10,12 +10,19 @@
 
 #include <vector>
 
+#include "memory_utils.h"
+
 struct TVMObject {
   tvm::PackedFunc kernel;
   tvm::PackedFunc set_input;
   tvm::PackedFunc get_output;
   // Map input indices to values in the subgraph
-  std::vector<torch::jit::Value*> input_values;
+  // Plus indicates if the corresponding value is immutable,
+  // e.g., a parameter such as weight.
+  std::vector<std::pair<torch::jit::Value*, bool>> input_values;
+  // DLManagedTensorPtr = unique_ptr<DLManagedTensor, DLManagedTensorDeleter>
+  std::unordered_map<torch::jit::Value*, torch_tvm::utils::DLManagedTensorPtr>
+    tvm_param_tensors;
 };
 
 struct TVMCompiler {
@@ -47,5 +54,5 @@ struct TVMCompiler {
   static tvm::relay::Function convertToRelay(
       std::shared_ptr<torch::jit::Graph> subgraph,
       TVMContext ctx,
-      std::vector<torch::jit::Value*>* input_values = nullptr);
+      std::vector<std::pair<torch::jit::Value*, bool>>* input_values = nullptr);
 };
--- a/torch_tvm/memory_utils.cpp
+++ b/torch_tvm/memory_utils.cpp
@@ -1,0 +1,58 @@
+#include <tvm/runtime/device_api.h>
+
+#include "memory_utils.h"
+
+namespace torch_tvm {
+namespace utils {
+
+bool isAligned(void* data_ptr, std::uintptr_t alignment_in_bytes) {
+  auto mask = alignment_in_bytes - 1;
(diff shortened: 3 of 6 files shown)


kimishpatel:
Depends on this PR: https://github.com/pytorch/tvm/pull/85

torch_tvm/operators.h
```diff
@@ -3,6 +3,12 @@
 #include <tvm/relay/expr.h>
 #include <tvm/relay/op.h>
 
+#define PARAM_INDICES_convolution {1, 2}
+#define PARAM_INDICES_layer_norm {2, 3}
+#define PARAM_INDICES_linear {1, 2}
+
+#define PARAM_INDICES(op_name) PARAM_INDICES_##op_name
```
bwasti: this is a bit strange, can we inline the {1, 2} etc into the list of ops as the final param?

torch_tvm/memory_utils.h
```diff
@@ -0,0 +1,37 @@
...
+namespace utils {
+
+struct DLManagedTensorDeleter {
+  void operator()(DLManagedTensor* manager_ctx) {
+    if (manager_ctx == nullptr)
+      return;
+    auto dl_tensor = manager_ctx->dl_tensor;
+    if (dl_tensor.data) {
+      delete dl_tensor.data;
+      // Implicit assumption is made that data, shape and strides are
```
bwasti: lets make this explicit with a CHECK

torch_tvm/memory_utils.cpp
```diff
@@ -0,0 +1,51 @@
...
+  auto num_dims = tensor.dim();
+  dl_tensor.ndim = num_dims;
+  dl_tensor.dtype = at::getDLDataType(tensor);
+  int64_t device_id = 0;
+  dl_tensor.ctx = getDLContext(tensor, device_id);
+  dl_tensor.shape = dl_tensor.strides = nullptr;
+  dl_tensor.data = nullptr;
+  dl_tensor.shape = new int64_t[num_dims];
+  dl_tensor.strides = new int64_t[num_dims];
+  AT_CHECK(dl_tensor.shape != nullptr && dl_tensor.strides != nullptr,
```
bwasti: AT_CHECK is deprecated, use TORCH_CHECK

torch_tvm/memory_utils.cpp
```diff
@@ -0,0 +1,51 @@
+#include "memory_utils.h"
+
+namespace torch_tvm {
+namespace utils {
+
+bool is_aligned(void* data_ptr, std::uintptr_t alignment_in_bytes) {
+  auto mask = alignment_in_bytes - 1;
+  AT_CHECK((alignment_in_bytes & mask) == 0);
```
bwasti: might be nice to make this a compile time check

torch_tvm/compiler.cpp
```diff
@@ -320,8 +342,22 @@ void TVMCompiler::run(Stack& stack) {
...
+    auto tensor = ivalue.toTensor();
+    DLManagedTensor* dl_tensor;
+    if (torch_tvm::utils::is_aligned(tensor.data_ptr(),
+          tvm::runtime::kAllocAlignment)) {
+      dl_tensor = at::toDLPack(tensor);
+    } else if (cache_[spec].tvm_param_tensors_.value_exists(value)) {
+      // Is the value parameter that is assumed to immutable?
+      dl_tensor = cache_[spec].tvm_param_tensors_.
+        get_tensor_like_for_value(value, tensor);
+    } else {
```
bwasti: isn't this final case unsound?

torch_tvm/compiler.cpp
```diff
@@ -320,8 +342,22 @@ void TVMCompiler::run(Stack& stack) {
...
-    auto tensor = ivalue.toTensor().to(at::kFloat);
-    auto dl_tensor = at::toDLPack(tensor);
+    //auto tensor = ivalue.toTensor().to(at::kFloat);
+    auto tensor = ivalue.toTensor();
+    DLManagedTensor* dl_tensor;
+    if (torch_tvm::utils::is_aligned(tensor.data_ptr(),
+          tvm::runtime::kAllocAlignment)) {
+      dl_tensor = at::toDLPack(tensor);
+    } else if (cache_[spec].tvm_param_tensors_.value_exists(value)) {
+      // Is the value parameter that is assumed to immutable?
```
bwasti: let's chuck in a warning here just in case

torch_tvm/compiler.h
```diff
@@ -10,12 +10,15 @@
...
 
+#include "managed_tensors.h"
+
 struct TVMObject {
   tvm::PackedFunc kernel;
   tvm::PackedFunc set_input;
   tvm::PackedFunc get_output;
   // Map input indices to values in the subgraph
   std::vector<torch::jit::Value*> input_values;
+  torch_tvm::ManagedParamTensors tvm_param_tensors_;
```
bwasti: nit: no trailing `_`, this is a public variable

torch_tvm/compiler.cpp
```diff
@@ -126,6 +129,7 @@ tvm::relay::Expr TVMCompiler::convertToRelay(
 tvm::relay::Function TVMCompiler::convertToRelay(
     std::shared_ptr<Graph> subgraph,
     TVMContext ctx,
+    torch_tvm::ManagedParamTensors* tvm_param_tensors_ptr,
```
bwasti: can we just use `std::vector<size_t>* immutable_input_indices`?  Adding a new data-structure should be avoided here

torch_tvm/compiler.cpp
```diff
@@ -148,7 +152,15 @@ tvm::relay::Function TVMCompiler::convertToRelay(
       auto uses = value->uses();
       for (const auto& use : uses) {
         tvm::Array<tvm::relay::Expr> relay_inputs;
+        // Things like prim::Return
+        // Should we be more explicit here?
+        // That only prim::Return should be skipped?
+        if (use.user->outputs().size() < 1) {
```
bwasti: what is this logic doing here?

torch_tvm/compiler.cpp
```diff
@@ -169,15 +181,18 @@ tvm::relay::Function TVMCompiler::convertToRelay(
               }
             }
           }
+          if (!skip_user &&
+              tvm_param_tensors_ptr &&
+              std::find(param_indices.begin(),
+              param_indices.end(), input_index) != param_indices.end()) {
+            tvm_param_tensors_ptr->insert_value(input);
```
bwasti: this is much easier to understand with a standard container instead of a custom data structure

torch_tvm/managed_tensors.h
```diff
@@ -0,0 +1,77 @@
...
+      auto managed_tensor_ptr = torch_tvm::utils::
+        alloc_and_copy_data(tensor);
+      managed_tensors[val] = DLManagedTensorPtr(managed_tensor_ptr);
+    }
+
+  protected:
+    std::unordered_map<const Value*, DLManagedTensorPtr> managed_tensors;
+};
+
+class ManagedParamTensors : public ManagedTensors {
```
bwasti: why is this needed?

torch_tvm/memory_utils.cpp
```diff
@@ -0,0 +1,51 @@
...
+  return (reinterpret_cast<std::uintptr_t>(data_ptr) & mask) == 0;
+}
+
+DLManagedTensor* alloc_and_copy_data(const at::Tensor& tensor) {
+  DLManagedTensor* dl_managed_tensor = new DLManagedTensor();
+  // managed_tensor_deleter is supplied to unique_ptr as a deleter
+  // of this managed memory. Thus setting deleter to nullptr;
+  dl_managed_tensor->deleter = nullptr;
```
bwasti: why are we hijacking the lifetime logic? can we hand over a deleter that works with DLManagedTensor?  we don't want to have to maintain this lifetime stuff

torch_tvm/operators.h
```diff
@@ -11,17 +17,23 @@ tvm::relay::Expr getOperator(
...
 uint64_t getNoneSentinel();
 
+const std::vector<int32_t>& getParamIndices(torch::jit::Node* node);
+
 using TVMOpFunctor = std::function<tvm::relay::Expr(
     torch::jit::Node* node,
     tvm::Array<tvm::relay::Expr> inputs)>;
 using TVMScheduleFunctor = std::function<const tvm::runtime::PackedFunc*()>;
 
+using ParamIndicesType = std::vector<int32_t>;
```
bwasti: no need for this, it's readable without it

torch_tvm/operators.cpp
```diff
@@ -216,7 +222,8 @@ RegisterTVMOperator reg({
              bias_add_op, {out, inputs[2]}, tvm::Attrs(bias_add_attrs), {});
        }
        return out;
-     }},
+     },
+     "", PARAM_INDICES(convolution)},
```
bwasti: re my comment below, this would be fine as
```
}, "", {1, 2}},
```

torch_tvm/operators.cpp
```diff
@@ -23,12 +23,18 @@ std::unordered_map<Symbol, TVMOpFunctor>& getTVMOperatorMap() {
   return map;
 }
 
+std::unordered_map<Symbol, ParamIndicesType>& getOpParamsMap() {
```
bwasti: I wonder if we should change getTVMOperatorMap to just return the `struct TVMOpMap` (and maybe even rename that to something a bit more understandable)

torch_tvm/compiler.cpp
kimishpatel: I just moved the logic up here. It was always there. We do not translate return to any tvm op. So we dont translated it.

torch_tvm/compiler.cpp
kimishpatel: We need this if the input data, not parameter, is unaligned.

torch_tvm/managed_tensors.h
kimishpatel: I was under the impression that the deleter had to be explicitly called and given we are creating this tensor who would call the deleter. But it seems NDArray does call the deleter of underlying DLManagedTensor if specified. So yes I agree, we can get rid of much of these wrappers.

kimishpatel: @bwasti ^^

torch_tvm/compiler.cpp
kimishpatel: Why is that? Note that in the current implementation this is needed. We cannot just get away with the indices.

torch_tvm/compiler.cpp
kimishpatel: This is inserting the Value pointer in the cache. Perhaps if `tvm_param_tensor_ptr` is names as `tvm_param_cache_ptr` it might be more readable.

torch_tvm/memory_utils.cpp
kimishpatel: Well if we constexpr alignment to 64 or some value then perhaps we can convert this to static_assert. Then rename this to is_64B_aligned. But the alignment requirement comes from tvm so even though we know it needs 64B aligned, making it constexpr here will decouple it from tvm. Unfortunately due to some include issues I had to hardcode that value to 64 in some places, but that I intend to fix.

torch_tvm/memory_utils.cpp
kimishpatel: Read my comment above. I think we will have to still maintain the cache of some sort. If we dont want to manage lifetime we can hand it over to NDArray the cache will require to map Value* to NDArray*. If we agree on that, I think that is doable.

torch_tvm/operators.cpp
kimishpatel: It might be better. I had to add `getOpParamsMap` and corresponding `param_map` because we dont save TVMOpMap. Also probably should rename `TVMOpMap` to `TVMOp`?

torch_tvm/operators.cpp
kimishpatel: So I dont like constant 1s and 2s lying around without explaining what they are. `PARAM_INDICES` conveys clearly what they are. If you think it is overly complicated to go somewhere else and add a define for a new op, I understand that. But note that it is cleaner this way. I would have `op_params.h` where all such defines exist. We can also make that more readable by having comments there.

torch_tvm/operators.h
kimishpatel: See my comment above as to why I chose to do this way.

torch_tvm/managed_tensors.h
```diff
@@ -0,0 +1,77 @@
...
+#include <tvm/runtime/device_api.h>
+
+#include <torch/csrc/jit/ir.h>
+#include <ATen/DLConvertor.h>
+
+#include "memory_utils.h"
+
+namespace torch_tvm {
+namespace {
+}
```
bwasti: rm

yinghai:
Fixes https://github.com/pytorch/tvm/issues/80

torch_tvm/memory_utils.cpp
yinghai: You can make `alignment_in_bytes` a template argument and use static_assert.

torch_tvm/operators.h
yinghai: Yeah.

torch_tvm/managed_tensors.h
yinghai: can't it just be `ManagedTensors`?

torch_tvm/managed_tensors.h
```diff
@@ -0,0 +1,77 @@
...
+
+using torch::jit::Value;
+
+using torch_tvm::utils::DLManagedTensorPtr;
+
+class ManagedTensors {
+  public:
+    ManagedTensors() = default;
+    bool insert_value(const Value* val) {
```
yinghai: BTW, generally we use `insertValue` kind of naming convention?

torch_tvm/compiler.cpp
```diff
@@ -320,8 +342,22 @@ void TVMCompiler::run(Stack& stack) {
...
     auto ivalue = value_to_ivalue.at(cache_[spec].input_values[i]);
-    auto tensor = ivalue.toTensor().to(at::kFloat);
-    auto dl_tensor = at::toDLPack(tensor);
+    //auto tensor = ivalue.toTensor().to(at::kFloat);
+    auto tensor = ivalue.toTensor();
+    DLManagedTensor* dl_tensor;
+    if (torch_tvm::utils::is_aligned(tensor.data_ptr(),
+          tvm::runtime::kAllocAlignment)) {
+      dl_tensor = at::toDLPack(tensor);
+    } else if (cache_[spec].tvm_param_tensors_.value_exists(value)) {
```
yinghai: Eventually we don't want to even compare with this value if we already know what's treated as weights in TVM as you probably will go though alter layout and have your own transformed weights.

kimishpatel: yes, but we cant couple this with whether weights get transformed. Transformation is optimization step and not always done or can be relied upon. Right? Unless I am missing something.

torch_tvm/managed_tensors.h
kimishpatel: Yes, removing this whole thing anyway.

torch_tvm/managed_tensors.h
kimishpatel: Yes, I will remove this and do it alternatively.

torch_tvm/memory_utils.cpp
kimishpatel: Only if kAllocAlignment value which comes from tvm is constexpr. I will check and see.

torch_tvm/compiler.cpp
yinghai: Once you enable opt_lvl=3, you cannot expect that your weights are not changed.

kimishpatel: I dont think I am following you. The only expectation above condition sets is that we created and copied tensor that corresponds to some Value*. Here that tensor is weight. What happens inside tvm, as a result opt_lvl, is not what this intends to do.
Can you elaborate what you mean by "you cannot expect that your weights are not changed"? Are you talking about layout transformation of weights?

yinghai: > Are you talking about layout transformation of weights?

Yes.

kimishpatel:
@yinghai, thats a good point. For that we will have to make the aligned DLTensor that we created part of the `cache_` that caches compilation, so that DLTensor is destructed when that is destructed. In another PR.

torch_tvm/compiler.cpp
kimishpatel: Only reason the check is done is because: at the time of relay graph creation, the only thing we have is Value*. But this is not strictly true. We create relay graph on the very first inference which means the weights are definitely available. I think we can have `std::pair<Value*, DLManagedTensorPtr>`, instead of `std::pair<Value*, bool>` that maps Value* to its mutability property. But it probably makes code little less readable. Thanks for the input.

PR #88, opened 2019-08-15
Fusion logic bug fix.

kimishpatel:
Earlier we would create a fusion group and jump onto the next node, once
no more nodes can be added to the current group. This leads to some
nodes of original graph being skipped and the resulting iterator, when
current group stops expanding, skips some nodes.
This fix now iterates over the graph until no fusions. The logic is
lifted straight out of graph_fuser.cpp in PyTorch graph passes.

The fusion part here is largely untested right now. Will need to add more
tests.

Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torch_tvm/fusion_pass.cpp
+++ b/torch_tvm/fusion_pass.cpp
@@ -101,7 +101,7 @@ c10::optional<Node*> tryMerge(
 }
 #undef REQ
 
-graph_node_list::iterator scanNode(
+std::pair<graph_node_list::iterator, bool> scanNode(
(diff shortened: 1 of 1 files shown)


zrphercule:
roger!

Yinghai Lu <notifications@github.com>于2019年8月19日 周一上午10:55写道：

kimishpatel:
> please rebase, let tests run, then merge :)

For sure :). Will rebase my PRs.

PR #89, opened 2019-08-16
Custom dense op

kimishpatel:
Pulling in changes dense op when the input has more than 2 dims. It essentially flattens first n-1 dims.


torch_tvm/custom_tvm_ops/custom_dense.py
```diff
@@ -0,0 +1,79 @@
...
+                bias, out_dtype, pack_weights)
+
+    # For small batch sizes, don't pack weight into cache-friendly layout
+    # because of overhead in packing and limited reuse from batch dimension
+    # TODO(icemelon9): use a more systematic way to determine which schedule to use
+    if batch <= 16:
+        return _declaration_dense_nopack(cfg, data, weight, bias, out_dtype)
+    return _declaration_dense_pack(cfg, data, weight, bias, out_dtype)
+
+@autotvm.register_topi_compute(nn.dense, "cpu", "custom_direct_reshape")
```
llyfacebook: I don't think we need this decorator.

test/test_operators.py
```diff
@@ -1,5 +1,5 @@
 import unittest
-from test.util import TVMTest
```
bwasti: what's going on here? I think this will break CI

kimishpatel: Strange. Will fix this.

torch_tvm/custom_tvm_ops/custom_dense.py
kimishpatel: No? Do you mean just use `tvm.register_topi_compute`? I am not sure but the function does define split on the axis which can be used for tuning.

kimishpatel:
@yinghai, this is just a copy paste of original implementation in tvm. The only thing it is supposed to do in flatten on the input and reshape on the output. Actually I will replace what seems like customized op with call to original if possible. Thanks for the input. Regarding benchmarking, we dont have yet. It is something that needs to be done. Also I dont think we can quite support any number of dims because torch.linear is more generous and handles many cases, so I am adding new checks.

yinghai:
> The only thing it is supposed to do in flatten on the input and reshape on the output.  

Can we just add a `flatten` symbol before the input when we construct the relay graph?

kimishpatel:
In the current implementation we may or may not have that information (i.e., the shape of the input) during relay graph construction.

bwasti:
If we have input sizes we should have the shape information associated with every intermediate, right?  We can deduce the input shape from the input expr with the type inference pass: https://bddppq.github.io/codebrowser/tvm/tvm/src/relay/pass/type_infer.cc.html#_ZN3tvm5relay9InferTypeERKNS0_4ExprERKNS0_6ModuleE

yinghai:
> In the current implementation we may or may not have that information (i.e., the shape of the input) during relay graph construction. 

This doesn't sound right. Why don't we have this piece of info? Do we only have info at the primary inputs of the graph? If so, we should consider getting more shape info.

kimishpatel:
Right now we construct relay graph on the first run of the JIT graph. We can postpone this to second run also and then should have ProfiledTensorType for each Value*. 
I am not sure how easy it is to run type inference pass though. From the quick look at signature `Expr InferType(const Expr& expr, const Module& mod_ref)` it seems that it needs Relay Module as well.

yinghai:
> In the current implementation we may or may not have that information (i.e., the shape of the input) during relay graph construction. 

Check @bwasti's code pointer. You can infer the shape incrementally in relay for now. I think eventually, we are going to the profiling based jit runner. But I don't think it's ready.

kimishpatel:
> > In the current implementation we may or may not have that information (i.e., the shape of the input) during relay graph construction.
> 
> Check @bwasti's code pointer. You can infer the shape incrementally in relay for now. I think eventually, we are going to the profiling based jit runner. But I don't think it's ready.

Thats what I am referring to my in my comment above. We can do this in a separate PR thought if we want to go that route.

PR #91, opened 2019-08-16
Convert PT Int to 64 bit

kimishpatel:
Summary:
PT ints are 64 bit integers. Without this type mismatch is causing
asserts on TVM side. This PR fixes that.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torch_tvm/compiler.cpp
+++ b/torch_tvm/compiler.cpp
@@ -98,8 +98,6 @@ tvm::relay::Expr TVMCompiler::convertToRelay(
   if (val.isInt()) {
     auto x = tvm::runtime::NDArray::Empty({}, tvm::Int(64), ctx);
     auto l = val.toInt();
-    TORCH_CHECK(l <= std::numeric_limits<int64_t>::max());
-    TORCH_CHECK(l >= std::numeric_limits<int64_t>::lowest());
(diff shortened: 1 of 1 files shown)


torch_tvm/compiler.cpp
```diff
@@ -109,11 +108,10 @@ tvm::relay::Expr TVMCompiler::convertToRelay(
   if (val.isIntList()) {
     tvm::Array<tvm::relay::Expr> tuple_elems;
     for (const auto& elem : val.toIntList()) {
-      auto x = tvm::runtime::NDArray::Empty(
-          {}, tvm::runtime::String2TVMType("int32"), ctx);
-      TORCH_CHECK(elem <= std::numeric_limits<int32_t>::max());
-      TORCH_CHECK(elem >= std::numeric_limits<int32_t>::lowest());
-      reinterpret_cast<int32_t*>(x->data)[0] = elem;
+      auto x = tvm::runtime::NDArray::Empty({}, tvm::Int(64), ctx);
+      TORCH_CHECK(elem <= std::numeric_limits<int64_t>::max());
```
bwasti: this shouldn't be necessary (elem is a signed 64bit int)

kimishpatel: Aah. makes sense. Too many things in the same day.

PR #94, opened 2019-08-20
Changed linear to handle more than 2 dims for input.

kimishpatel:
Done via invoking type inference.
Unit tested.

Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torch_tvm/operators.cpp
+++ b/torch_tvm/operators.cpp
@@ -1,5 +1,7 @@
 #include <tvm/relay/attrs/nn.h>
 #include <tvm/relay/attrs/transform.h>
+#include <tvm/relay/transform.h>
+#include <tvm/relay/module.h>
 
 #include <torch/csrc/autograd/record_function.h>
 #include <torch/csrc/jit/custom_operator.h>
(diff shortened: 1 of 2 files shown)


kimishpatel:
Closed the other PR for dense op and replaced with this one.

torch_tvm/operators.cpp
```diff
@@ -13,6 +15,89 @@
...
+    uint64_t num_elements_to_flatten = 1;
+    int64_t i = 0;
+    for (;i < shape0.size()-1; ++i) {
+      auto d = (shape0[i].as<tvm::IntImm>())->value;
+      num_elements_to_flatten *= d;
+    }
+    auto last_dim = (shape0[i].as<tvm::IntImm>())->value;
+    auto attrs = tvm::make_node<tvm::relay::ReshapeAttrs>();
+    auto& newshape = attrs->newshape;
+    newshape.push_back(tvm::Integer(num_elements_to_flatten));
```
yinghai: I remember you can use `-1` but being explicit is good here.

torch_tvm/operators.cpp
```diff
@@ -505,30 +590,7 @@ RegisterTVMOperator reg({
...
-
-       if (!relayIsNone(inputs[2])) {
-         auto bias_add_op = tvm::relay::Op::Get("nn.bias_add");
-         auto bias_add_attrs = tvm::make_node<tvm::relay::BiasAddAttrs>();
-         bias_add_attrs->axis = 1;
-         return tvm::relay::CallNode::make(
-             bias_add_op, {out, inputs[2]}, tvm::Attrs(bias_add_attrs), {});
-       }
-       return out;
+       return lowerLinear(inputs);
```
yinghai: Nice. We should avoid having a lot of long anonymous functions.

PR #95, opened 2019-08-20
Map batch matmul from PT to TVM.

kimishpatel:
Also added test.

Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torch_tvm/operators.cpp
+++ b/torch_tvm/operators.cpp
@@ -503,6 +503,26 @@ RegisterTVMOperator reg({
            tvm::relay::CallNode::make(op, {inputs[0]}, tvm::Attrs(attrs), {});
        return out;
      }},
+    {Symbol::fromQualString("aten::bmm"),
+     [](Node* node, tvm::Array<tvm::relay::Expr> inputs) {
+       TORCH_INTERNAL_ASSERT(inputs.size()==2);
+
+       auto transpose_attrs = tvm::make_node<tvm::relay::TransposeAttrs>();
+       auto& axes = transpose_attrs->axes;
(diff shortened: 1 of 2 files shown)


torch_tvm/operators.cpp
```diff
@@ -503,6 +503,26 @@ RegisterTVMOperator reg({
            tvm::relay::CallNode::make(op, {inputs[0]}, tvm::Attrs(attrs), {});
        return out;
      }},
+    {Symbol::fromQualString("aten::bmm"),
+     [](Node* node, tvm::Array<tvm::relay::Expr> inputs) {
+       TORCH_INTERNAL_ASSERT(inputs.size()==2);
+
+       auto transpose_attrs = tvm::make_node<tvm::relay::TransposeAttrs>();
+       auto& axes = transpose_attrs->axes;
+       axes.push_back(0);
```
yinghai: Why this specific order?

yinghai (approved):

yinghai:
Regarding the error: https://stackoverflow.com/questions/34819221/why-is-python-setup-py-saying-invalid-command-bdist-wheel-on-travis-ci

kimishpatel:
Hmm, let me try with updating ci config. Thanks for the link.

yinghai:
Hmm, try rebasing and it should work.

yinghai:
Rebase to master and we should expect CI to pass.

kimishpatel:
This one is causing some correctness issues when run with bert model, even though unit tests pass. So I need to do some investigation before pushing it in.

PR #96, opened 2019-08-20
Master broken due to changed in PT API. Catching up with that.

kimishpatel:
Summary:
ProfiledTensorType's sizes returns optional<ListofInts>. Changes to
reflect that.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


torch_tvm/compiler.cpp
```diff
@@ -55,7 +55,10 @@ tvm::relay::Var TVMCompiler::convertToRelay(Value* val, TVMContext ctx) {
       "Expected CPU device type but got:", device_type);
     tvm::Array<tvm::relay::IndexExpr> sizes;
     const auto& varying_sizes = pt_t->sizes();
-    for (const auto& optional_size : varying_sizes.sizes()) {
```
yinghai: Yeah, it's weird that I had met with `use out of scope` ASAN error if I just chain the calls together.

PR #103, opened 2019-08-22
Set input bug fix

kimishpatel:
set_input_zero_copy is invoked via index. However order of inputs is not necessarily preserved after build when module is built using graph_json.

### DIFF
--- a/torch_tvm/compiler.cpp
+++ b/torch_tvm/compiler.cpp
@@ -135,7 +135,7 @@ tvm::relay::Expr TVMCompiler::convertToRelay(
 tvm::relay::Function TVMCompiler::convertToRelay(
     std::shared_ptr<Graph> subgraph,
     TVMContext ctx,
-    std::vector<std::pair<Value*, bool>>* input_values) {
+    std::unordered_map<torch::jit::Value*, TVMGraphInputInfo>* input_values) {
   std::unordered_map<Value*, tvm::relay::Expr> value_map;
   tvm::Array<tvm::relay::Var> input_vars;
 
@@ -145,7 +145,10 @@ tvm::relay::Function TVMCompiler::convertToRelay(
     input_vars.push_back(v);
     if (input_values) {
       // Primary inputs are always mutable.
-      input_values->emplace_back(input, false);
+      input_values->emplace(std::piecewise_construct,
+          std::forward_as_tuple(input),
+          std::forward_as_tuple(false,
+            v.as<tvm::relay::VarNode>()->name_hint()));
     }
     value_map[input] = v;
   }
@@ -180,12 +183,15 @@ tvm::relay::Function TVMCompiler::convertToRelay(
               break;
             } else {
               if (optional_ivalue.value().isTensor()) {
-                if (input_values) {
-                  input_values->emplace_back(input, false);
-                }
                 auto input_var = convertToRelay(input, ctx);
                 input_vars.push_back(input_var);
                 value_map[input] = input_var;
+                if (input_values) {
+                  input_values->emplace(std::piecewise_construct,
+                      std::forward_as_tuple(input),
+                      std::forward_as_tuple(false,
+                        input_var.as<tvm::relay::VarNode>()->name_hint()));
+                }
               } else {
                 value_map[input] = convertToRelay(optional_ivalue.value(), ctx);
               }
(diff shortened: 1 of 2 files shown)


torch_tvm/compiler.h
```diff
@@ -12,17 +12,29 @@
 
 #include "memory_utils.h"
 
+struct TVMGraphInputInfo {
+  TVMGraphInputInfo(bool is_param_, std::string tvm_var_name_) {
```
bwasti: Any reason this can’t just be a tuple?

kimishpatel: It makes code less readable. I dont think doing std::get<0,1> etc, helps understand what the parameters are.

kimishpatel: I did think about this fwiw.

torch_tvm/compiler.h
```diff
@@ -12,17 +12,29 @@
...
+    is_param = is_param_;
+    tvm_var_name = std::move(tvm_var_name_);
+  }
+  TVMGraphInputInfo(bool is_param_, std::string&& tvm_var_name_) {
+    is_param = is_param_;
+    tvm_var_name = tvm_var_name_;
+  }
+  std::string tvm_var_name;
+  bool is_param;
+  // DLManagedTensorPtr = unique_ptr<DLManagedTensor, DLManagedTensorDeleter>
```
bwasti: rm

torch_tvm/none_var_expr.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
```
bwasti: I wonder if this stuff should live in a tvm folder with the ops, as it extends tvm and can be upsteamed

kimishpatel: Yes. I was thinking the same. I will move it to `tvm` folder.

kimishpatel:
Removed NoneVarNode stuff as it seem little more involved than what I had done. Will file issue for future fix.

torch_tvm/compiler.h
yinghai: string && ?

torch_tvm/compiler.cpp
```diff
@@ -352,37 +358,37 @@ void TVMCompiler::run(Stack& stack) {
...
       auto optional_ivalue = toIValue(value);
       TORCH_INTERNAL_ASSERT(optional_ivalue.has_value());
       value_to_ivalue[value] = optional_ivalue.value();
     }
-    auto ivalue = value_to_ivalue.at(cache_[spec].input_values[i].first);
+    auto ivalue = value_to_ivalue.at(value);
     //auto tensor = ivalue.toTensor().to(at::kFloat);
     auto tensor = ivalue.toTensor();
     DLManagedTensor* dl_tensor;
     if (tensor.is_contiguous() &&
```
yinghai: Can we make a function out of this chunk?

torch_tvm/compiler.cpp
```diff
@@ -352,37 +358,37 @@ void TVMCompiler::run(Stack& stack) {
...
+        graph_input.tvm_tensor = DLManagedTensorPtr(dl_tensor);
       }
     } else {
       dl_tensor =
         torch_tvm::utils::allocAndCopyData(tensor);
       dl_tensor_list.emplace_back(
           dl_tensor);
     }
-    cache_[spec].set_input(i, tvm::runtime::NDArray::FromDLPack(dl_tensor));
+    cache_[spec].set_input(graph_input.tvm_var_name,
```
yinghai: I'd keep a map of IValue* and int. And set input by position. Hash a string is slower than hash a pointer.

torch_tvm/compiler.cpp
kimishpatel: Let me see if this is possible to do. Position of the node for a given variable name is known by the runtime. I am not sure if this is exposed, but I will look into it.

torch_tvm/compiler.h
kimishpatel: @yinghai, there is a constructor with string&&. A separate one.

torch_tvm/compiler.cpp
kimishpatel: I think the position of the nodes is known to runtime only. Internally it just does a search to find the index. I will see if it exposes indexed node names.

kimishpatel: Let me see if this is possible to do. Position of the node for a given variable name is known by the runtime. I am not sure if this is exposed, but I will look into it.

torch_tvm/compiler.cpp
kimishpatel: Sure.

torch_tvm/compiler.cpp
yinghai: OK. It's fine for this diff I think.

torch_tvm/compiler.cpp
```diff
@@ -196,14 +202,14 @@ tvm::relay::Function TVMCompiler::convertToRelay(
           if (!skip_user && input_values &&
               std::find(param_indices.begin(),
                 param_indices.end(), input_index) != param_indices.end()) {
-            auto it = std::find_if(input_values->begin(), input_values->end(),
-                [input](const std::pair<Value*, bool>& v)
-                {return v.first == input;});
-            // It is possible that parameter is not in input_values.
-            // This can happen for example when bias input is None in
-            // convolution.
+            /*
```
yinghai: ?

PR #105, opened 2019-08-23
Support clamp

bertmaher:


torch_tvm/operators.cpp
```diff
@@ -444,6 +444,16 @@ RegisterTVMOperator reg({
        auto out = tvm::relay::CallNode::make(op, inputs, tvm::Attrs(), {});
        return out;
      }},
+    {Symbol::fromQualString("aten::clamp"),
+     [](Node* node, tvm::Array<tvm::relay::Expr> inputs) {
+       auto op = tvm::relay::Op::Get("clip");
+       auto clip_attrs = tvm::make_node<tvm::relay::ClipAttrs>();
+       clip_attrs->a_min = relayToConstant<float>(inputs[1]);
```
yinghai: How do we know the second and third input is constant?

kimishpatel: At this point all scalars are make into `tvm::relay::ConstantNode` and I dont think they are part of `ArgumentSpec`, although I am not sure. So you can argue that two calls of `aten::clamp` can come with two different values of clamp and we certainly won't catch that.
But in the JIT graph you can check if those inputs come from `prim::Constant` nodes or not. That would be a good thing to add.

yinghai: @kimishpatel It makes more sense to me. Do we actually do this for other ops?

kimishpatel: we do `relayToConstant` everywhere, so it would be good to add this test. I will put it up for TODO.

bertmaher: Should this diff depend on adding those `prim::Constant` checks, then?

kimishpatel: I dont think so.

bertmaher:
Thanks for the review!  Can you merge (or give me write access so I can)? :)

kimishpatel:
Not sure how to give you the write access (I woke up one day and I had the write access :)), but I will merge.

PR #106, opened 2019-08-23
Remove dropout

kimishpatel:
Alteration pass to remove dropout nodes when training is false.
Existing mapping of dropout node is faulty and has correctness issues. Since removing dropout node when training=false can benefit PT as well, just decided to implement that instead.
Summary:

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torch_tvm/remove_dropout.h
+++ b/torch_tvm/remove_dropout.h
@@ -1,0 +1,5 @@
+#pragma once
+
+#include <torch/csrc/jit/ir.h>
+
+TORCH_API void RemoveDropout(std::shared_ptr<torch::jit::Graph>& graph);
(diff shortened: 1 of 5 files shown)


torch_tvm/remove_dropout.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+#include <torch/csrc/jit/ir.h>
+
+TORCH_API void RemoveDropout(std::shared_ptr<torch::jit::Graph>& graph);
```
yinghai: Feels like too specific to be a standalone file.

kimishpatel: True. But as of now it cannot quite merge into some other files. Maybe I can do generic graph_utils or something like that?

bwasti: seems fine for now (fuse linear does the same thing)

maybe we can do an optimize.h header for all of these in another diff

PR #107, opened 2019-08-23
Set parameter inputs of the graph.

kimishpatel:
Summary:
Setting parameter inputs enables constant propagation in TVM.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torch_tvm/compiler.cpp
+++ b/torch_tvm/compiler.cpp
@@ -6,6 +6,7 @@
 #include <torch/csrc/jit/interpreter.h>
 #include <limits>
 #include <tvm/runtime/device_api.h>
+#include <tvm/node/container.h>
 
 using namespace torch::jit;
 
@@ -19,6 +20,9 @@ std::vector<DLManagedTensorPtr> set_input(
   for (auto& input_value : cache.input_values) {
     auto* value = input_value.first;
     TVMGraphInputInfo& graph_input = input_value.second;
+    if (graph_input.is_param) {
+      continue;
+    }
     if (!value_to_ivalue.count(value)) {
       auto optional_ivalue = toIValue(value);
       TORCH_INTERNAL_ASSERT(optional_ivalue.has_value());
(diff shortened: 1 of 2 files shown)


torch_tvm/compiler.cpp
```diff
@@ -382,6 +422,15 @@ void TVMCompiler::run(Stack& stack) {
...
     cache_[spec].get_output = run_mod.GetFunction("get_output", false);
     auto get_num_outputs = run_mod.GetFunction("get_num_outputs", false);
+
+    // Set parameter inputs.
+    tvm::Map<std::string, tvm::relay::Constant> params = get_params();
+    for (const auto& param : params) {
+        const auto& param_name = param.first;
+        const auto& param_ndarray_val = param.second->data;
+        cache_[spec].set_input(param_name, param_ndarray_val);
+    }
```
llyfacebook: Are we setting the input value for params node twice? Assume we have done it before the build, right?

llyfacebook: Also this function is too large, I think we had better to refactor it.

kimishpatel: Yes the function is growing. Although I just did a refactor by moving set_input into a separate function. But regarding your question on setting params twice. We need to do this because the first time before build it is just to tell what are the param nodes. I am not sure at this point if it creates only the precompute net or also populates the weight value. I did try not setting it again and it was resulting in wrong value. After the build process, we need to set the values again but only once. This is certainly not efficient and can be improved upon on the tvm side, I suppose.

llyfacebook: hmm, If that's the case, how can we do the pre-compute pass in the building process. I think we need the concrete value.

kimishpatel: Well for my case it probably was necessary because there is nothing that you can precompute. Since the weight packing stuff is inside topi op it cannot be precomputed so I still need to provide the relevant data pointer.

kimishpatel: But either way, I dont think setting it again will mess with precompute if any could be done.

llyfacebook: I guess there might be some bug/issue for constant node.

PR #120, opened 2019-09-22
Quantized linear opt

kimishpatel:
Tensorize quantized linear.
Fixed a bug in the compute.
Depends on this PR.
https://github.com/facebookexperimental/tvm/pull/7

Specifically this PR does:
- Adds tensorization of compute (depends on the PR mentioned above) for AVX512.
- Reshape of input/output to handle input with > 2 dims.
- Weight reshape that is dependent on 1st dim of weight. That is if weight.shape[0] > 16 we expect shape size to be multiple of 16. Then we pack weight has NK16n4k. If weight.shape[0] < 16 we pack if as NK"weight.shape[0]n4k.
- Bug fixes:
  - In the quantized gemm calculation
```
result = lhs_scale * rhs_scale * Sum_over_i(
        (lhs_quantized_value[i] - lhs_zero_point) *
        (rhs_quantized_value[i] - rhs_zero_point)
    )
```
if you expland you get the last term as
`lhs_zero_point*rhs_zero_point*k`. 
However in fbgemm they do not have this addition. It is not clear yet why or something else is missed. We will followup on this. So this is one bug that is fixed.
- The second bug is more subtle. In input data quantization fbgemm uses rounding whereas we did casting. Casting by default does truncation. So whenever our quantized value is different it is always smaller by 1. When k dim is larger these get accumulated and with scale applied difference gets even larger. The solution was to use `tvm::round`, however this costs us 10% in perf. Debug is under way.

Benchmark comparision is against fbgemm implementation.
Without the rounding we are 10% faster with rounding we are on par. Need to figure out why are we losing perf due to rounding.

### DIFF
--- a/test/test_operators.py
+++ b/test/test_operators.py
@@ -340,10 +340,10 @@ class TestOperators(TVMTest):
             torch.fbgemm_linear_quantize_weight(weight.clone().float())
         packed_weight = torch.fbgemm_pack_quantized_matrix(q_weight.clone())
 
-        def fbgemm_quantized_linear(input, weight, bias):
+        def fbgemm_quantized_linear(input, weight, bias, col_offsets):
             return torch.fbgemm_linear_int8_weight_fp32_activation(
-                input.float(), q_weight, packed_weight, col_offsets, scale, zero_point, bias.float())
-        ref_out, tvm_out = self.runBoth(fbgemm_quantized_linear, input, weight, bias)
+                input.float(), weight, packed_weight, col_offsets, scale, zero_point, bias.float())
+        ref_out, tvm_out = self.runBoth(fbgemm_quantized_linear, input, weight, bias, col_offsets)
         # relax the constraint to avoid flaky test
         assert torch.allclose(ref_out, tvm_out, rtol=0.5, atol=0.5)
 
--- a/test/test_operators.py
+++ b/test/test_operators.py
@@ -328,8 +328,10 @@ class TestOperators(TVMTest):
         assert torch.allclose(ref_out_no_bias, tvm_out_no_bias, rtol=0.01, atol=0.01)
 
 
+    # Have to make min_dim large since we may compare non tensorized imlementation
+    # against fbgemm and that requires K dim to be large enough.
     @TVMTest.given(
-        shape=TVMTest.rand_shape(rank=2, min_dim=10),
+        shape=TVMTest.rand_shape(rank=2, min_dim=32, max_dim=64),
         out_features=TVMTest.rand_int(15, 64),
     )
     def test_quantized_linear(self, shape, out_features):
@@ -343,9 +345,9 @@ class TestOperators(TVMTest):
         shape[1] = shape[1] * 4
         if out_features > 16:
             out_features = out_features * 16
-        input = torch.rand(shape)
-        weight = torch.rand(out_features, shape[1])
-        bias = torch.rand(out_features)
+        input = torch.normal(torch.rand(shape))
+        weight = torch.normal(torch.rand(out_features, shape[1]))
+        bias = torch.normal(torch.rand(out_features))
         q_weight, col_offsets, scale, zero_point = \
             torch.fbgemm_linear_quantize_weight(weight.clone().float())
         packed_weight = torch.fbgemm_pack_quantized_matrix(q_weight.clone())
--- a/test/test_operators.py
+++ b/test/test_operators.py
@@ -1,5 +1,5 @@
 import unittest
-from test.util import TVMTest
+from util import TVMTest
 from torch.testing import FileCheck
 
 import torch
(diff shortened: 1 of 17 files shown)


kimishpatel:
@yinghai, updatd.

test/test_operators.py
```diff
@@ -330,21 +330,34 @@ def linear_no_bias(input, weight):
...
         shape=TVMTest.rand_shape(rank=2, min_dim=10),
-        out_features=TVMTest.rand_int(15, 30),
+        out_features=TVMTest.rand_int(15, 64),
     )
     def test_quantized_linear(self, shape, out_features):
+        # This is necessary since for N of size > 16 we enforce it to be
+        # multiple of 16. Right this defines packing of weights and that
+        # specifically makes this requirement necessary.
+        # On the other hand loosing this requirement needs changes that require
+        # some changes on relay side which otherwise complains during shape
```
llyfacebook: Can you elaborate the change? Or point out to some gist?

torch_tvm/custom_tvm_ops/cpp/relay/quantize_init.cc
```diff
@@ -35,6 +38,16 @@ RELAY_REGISTER_OP("nn.quantize_data_int8_quantize")
...
 
 
+RELAY_REGISTER_OP("nn.quantize_data_int8_row_offset")
+.describe(R"code(dynamic row offset calculation of quantized data.
+- **data**: (M, N)
+)code" TVM_ADD_FILELINE)
+  .set_num_inputs(1)
+  .add_argument("data", "Tensor", "Quantized input tensor.")
+  .set_support_level(10)
+  .add_type_rel("DataInt8RowOffset", DataInt8RowOffsetRel);
```
llyfacebook: In terms of return multiple tensors, can you check "vision.get_valid_counts" in tvm repo?

torch_tvm/custom_tvm_ops/cpp/topi/contrib/quantize.cc
```diff
@@ -20,18 +20,31 @@ TVM_REGISTER_GLOBAL("tvm.contrib.find_minmax")
...
+      int num_left_overs = num_els % 16;
+      float min_v[16] = {0.f};
+      float max_v[16] = {0.f};
       for (int i = 0; i < m; i++) {
-        for (int j = 0; j < n; j++) {
-            d_min = std::min(data_ptr[i*n + j], d_min);
-            d_max = std::max(data_ptr[i*n + j], d_max);
+        for (int j = 0; j < 16; j++) {
+            min_v[j] = std::min(data_ptr[i*16 + j], min_v[j]);
+            max_v[j] = std::max(data_ptr[i*16 + j], max_v[j]);
```
llyfacebook: Can you give a slow path for n less than 16?

test/test_operators.py
kimishpatel: In order to support arbitrary size on N we need to do:
1. Fix TVM. I will put up a "DONT COMMIT" PR for this. The reason we need to make this change is that right now if you try to do layout_transform in a way that dimension are not divisible, then things break because such transforms are not supported. It will silently do something else probably. For example.
Say you have weights: [10, 4]. layout_tranform from NK to NK16n4k means your new tensor is [1, 1, 16, 4]. Here dim[2] of 16 will have to be packed with zero for invalid elements. This does not happen today.
2. The REAL problem is this: Say we have a gemm of [1, 4]x[10,4]=[1,10]. Now after layout transform we will have our gemm produce [1, 16]. Post this we do requant. I rewrote requant to do take in [1, 16] and produce [1, 10] so that it only covers part of the tensor. In our quantized gemm implementation gemm and requant are part of the same op. Due to this relay complains during shape propagation that [1, 16] is larger than [1, 10]. A solution to this would be to split them up. We did not go that far and instead decided to rely on this assumption for now since for bert model most shapes are conformant.

kimishpatel: @llyfacebook, is that the answer you were looking for? Or your question was about something else?

torch_tvm/custom_tvm_ops/cpp/topi/contrib/quantize.cc
kimishpatel: This covers should cover all shapes. However I need to initialize min and max appropriately. 0 is not the right value for those.

torch_tvm/register.cpp
```diff
@@ -34,7 +34,7 @@ PYBIND11_MODULE(_torch_tvm, m) {
   options.setAliasAnalysis(AliasAnalysisKind::PURE_FUNCTION);
   RegisterOperators op({Operator(
       getTVMSymbol(),
-      [](const Node* node) {
+      [](const Node* node) -> Operation {
```
serhaty: since this is breaking build, can you please split this one and land separately? It also makes review easier by not merging unrelated changes in same PR.

kimishpatel: Yes. I am just trying to move faster since I will out on leave soon. But if you feel strongly about it, I can split it out.

torch_tvm/custom_tvm_ops/cpp/relay/quantize_init.cc
kimishpatel: @llyfacebook, so I checked that. The difference really is because of `@autotvm.register_topi_compute(data_int8_quantize, 'cpu', ['direct'])`. It is the `register_topi_compute` of autotvm that breaks. If we just go with generic_func like how they do it for get_valid_counts, it would be ok, but I did not try that. Maybe we can do that, although I am not too sure if that is going to buy us any advantage at the moment, because it seems unlikely that we can fuse `data quantize` and `row offset calc` as of now. (Mainly because right now you cannot inline primary outputs of an op)

torch_tvm/custom_tvm_ops/cpp/topi/contrib/quantize.cc
kimishpatel: @llyfacebook, fixed the bug.

PR #122, opened 2019-09-23
[torch_tvm] Support Lowering to TVM even if node cannot be fused

serhaty:
We take PT JIT IR graph and fuse those nodes of the graph that can be lowered to Relay graph in TVM. However the current fusion logic does not create a subgraph out of a single node. You need at least two nodes that can be lowered and only then it will fuse them and create a subgraph.

In this PR, we want to enable subgraph creation for single nodes that cannot be fused with any other node. We achieve this by checking for the case where consumer can be lowered but producer cannot be.

### DIFF
--- a/test/test_core.py
+++ b/test/test_core.py
@@ -2,6 +2,7 @@ import unittest
 from test.util import TVMTest
 import torch
 import torch_tvm
+import torch.nn.functional as F
 
 
 class TestCore(TVMTest):
@@ -134,6 +135,33 @@ class TestCore(TVMTest):
         assert "aten::dropout" not in \
                 str(tvm_graph_inference.graph_for(input_a, input_b, input_c)), \
                 "dropout must be removed during inference."
+
+    @TVMTest.given(
+        shape=TVMTest.rand_shape(rank=2, min_dim=4),
+        out_features=TVMTest.rand_int(3, 6),
+    )
+    def test_fuse_single_node(self, shape, out_features):
+        print("Running test for test_fuse_single_node")
+        input = torch.rand(shape)
+        weight = torch.rand(out_features, shape[1])
+        bias = torch.rand(out_features)
+
+        # check single node graph
+        def linear(a, b, c):
+            return F.linear(a, b, c)
+
+        ref_out, tvm_out = self.runBoth(linear, input, weight, bias)
+        assert torch.allclose(ref_out, tvm_out, rtol=0.01, atol=0.01)
+
+         # check to verify fusion still works
+        def linearSum(a, b, c):
+            return F.linear(a, b, c) + 2.0
+
+        ref_out, tvm_out = self.runBoth(linearSum, input, weight, bias)
+        assert torch.allclose(ref_out, tvm_out, rtol=0.01, atol=0.01)
+
+
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/torch_tvm/fusion_pass.cpp
+++ b/torch_tvm/fusion_pass.cpp
@@ -48,6 +48,21 @@ bool canHandle(Block* block, AliasDb& aliasDb) {
     GRAPH_DEBUG("Failed cond " #cond "\n"); \
     return c10::nullopt;                    \
   }
+
+c10::optional<Node*> tryLower(Node* node, AliasDb& aliasDb) {
+  GRAPH_DEBUG("Trying to lower node ", node->kind().toQualString(), ":\n");
+  // Already converted so return no change
+  REQ(node->kind() != getTVMSymbol() && !node->hasAttribute(attr::Subgraph));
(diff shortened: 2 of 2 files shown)


torch_tvm/fusion_pass.cpp
```diff
@@ -59,10 +59,26 @@ c10::optional<Node*> tryMerge(
       consumer->kind().toQualString(),
       ":\n");
 
-  // Symbolic checks
-  REQ(canHandle(producer, aliasDb));
+  // Symbolic check
   REQ((canHandle(consumer, aliasDb) || consumer->kind() == getTVMSymbol()));
 
+  // if producer cannot be converted, check if consumer can be lowered to TVM
```
kimishpatel: I think we should do this outside of this function. `tryMerge` is supposed to merge two nodes into a lowerable node. Iideally we start with a seed outside. Maybe 
```
if (canLowerSeed(consumer
```
If seed cannnot be lowered, we never come to `tryMerge`?

Also @bwasti, maybe you should chime in here.

test/test_core.py
```diff
@@ -135,5 +136,32 @@ def dropout_inference(a, b, c):
...
+        bias = torch.rand(out_features)
+
+        # check single node graph
+        def linear(a, b, c):
+            return F.linear(a, b, c)
+
+        ref_out, tvm_out = self.runBoth(linear, input, weight, bias)
+        assert torch.allclose(ref_out, tvm_out, rtol=0.01, atol=0.01)
+
+         # check to verify fustion still works
```
kimishpatel: spell check on 'fusion'.

test/test_core.py
```diff
@@ -135,5 +136,32 @@ def dropout_inference(a, b, c):
...
+    )
+    def test_fuse_single_node(self, shape, out_features):
+        print("Running test for test_fuse_single_node")
+        input = torch.rand(shape)
+        weight = torch.rand(out_features, shape[1])
+        bias = torch.rand(out_features)
+
+        # check single node graph
+        def linear(a, b, c):
+            return F.linear(a, b, c)
```
kimishpatel: Did you check if the test fails without your changes?

kimishpatel (changes requested):
Left a couple of comments. Also asked @bwasti to chime in.

kimishpatel:
Also maybe we can add an flag to enable this feature or not? Or maybe just check if it affects performance?

torch_tvm/fusion_pass.cpp
yinghai: Yeah, I think we should put this kind of logic in parallel to tryMerge instead of inside. I actually think we should have some kind of flag to control minimal number of ops in order to create a fusion group. But this might not be the scope of the PR.

torch_tvm/fusion_pass.cpp
```diff
@@ -59,10 +59,26 @@ c10::optional<Node*> tryMerge(
...
+  if(!canHandle(producer, aliasDb)){
+    if(consumer->kind() == getTVMSymbol() || consumer->hasAttribute(attr::Subgraph)) {
+      // Already converted so return no change
+      return c10::nullopt;
+    }
+    // prooceed to convert current node to TVM
```
yinghai: Typo

test/test_core.py
serhaty: yeap, it does indeed crash for single op version. For fusion one, I added to make sure future changes do not break it.

torch_tvm/fusion_pass.cpp
serhaty: I initially implemented as two passes:
1) Lower all nodes that can be lowered
2) Do Fusion among nodes that are already lowered

1 works fine but fusing two tvm nodes does not quire work as is.

torch_tvm/fusion_pass.cpp
```diff
@@ -59,9 +63,22 @@ c10::optional<Node*> tryMerge(
...
+  if(!canHandle(producer, aliasDb)){
+    if(consumer->kind() == getTVMSymbol() || consumer->hasAttribute(attr::Subgraph)) {
+      // Already converted so return no change
+      return c10::nullopt;
+    }
+    // proceed to convert current node to TVM
+    if(!aliasDb.isMutable(consumer)){
+      REQ(!aliasDb.hasOutputWriters(consumer));
+    }
+    consumer = SubgraphUtils::createSingletonSubgraph(consumer, getTVMSymbol());
```
kimishpatel: I also meant to suggest that we not only check if it can be lowered, but if it can be we create the singleton graph outside. Thus tryMerge always has consumer that is already lowered and it only does merging but never creation.
I think it will be cleaner this way. Sorry if it feels like I am dragging much want to drag it much.

llyfacebook:
Can you try the flow on the quantized model to see how many subgraph we could get now?

PR #123, opened 2019-09-24
To aid debugging of TVM lowered graph

kimishpatel:


torch_tvm/fusion_pass.cpp
```diff
@@ -24,7 +24,7 @@ value_list sortReverseTopological(ArrayRef<Value*> inputs, Block* block) {
 bool canHandle(Block* block, AliasDb& aliasDb);
 bool canHandle(Node* node, AliasDb& aliasDb) {
   if (node->kind() == prim::Constant) {
-    return true;
+    return false;
```
yinghai: Why this?

torch_tvm/compiler.cpp
```diff
@@ -385,6 +391,13 @@ void TVMCompiler::run(Stack& stack) {
             *subgraph_);
       }
     }
+
+    if (debug_) {
+      std::cout <<"subgraph \n";
```
yinghai: Saving to a file might be better.

torch_tvm/fusion_pass.cpp
kimishpatel: Aah I thought I removed this. Let me double check. Good catch.

torch_tvm/compiler.cpp
kimishpatel: Sure. Let me make changes.

torch_tvm/fusion_pass.cpp
kimishpatel: OK, had forgotten push the latest.

torch_tvm/compiler.cpp
```diff
@@ -419,14 +450,49 @@ void TVMCompiler::run(Stack& stack) {
     // This help performance.
     build_config->partition_const_loop = true;
     build_f(tvm_func, target_map, tvm::Target::Create(host_));
-    std::string json = json_f();
     tvm::runtime::Module mod = mod_f();
+    std::string json = json_f();
+    if (debug_) {
```
yinghai: Maybe we should just use a class, say DebugLogger to isolate these code.

kimishpatel: Let me think.

kimishpatel:
@yinghai, made changes to consolidate debug stuff. Is that what you had in mind? Also note that this is not a functional debug, this is for perf stuff. I mostly used it to see the schedule tvm used and the quality of the generated code. Was immensely helpful to me in perf debug, but wanted to clarify that this is not for functional debug.

torch_tvm/register.cpp
```diff
@@ -53,6 +55,11 @@ PYBIND11_MODULE(_torch_tvm, m) {
       FuseConcat(g);
       RemoveDropout(g);
       FuseSupportedOps(g);
+      if (debug) {
```
yinghai: Can this get into `DebugLogger` too?

kimishpatel: Let me think. Might be a bit tricky, but would be good to put alltogether.

kimishpatel:
BTW, it needs this PR to land too, else we will get runtime error (I need to fix that so that we dont core dump or something). https://github.com/facebookexperimental/tvm/pull/9. Do you want me to push that in as well?

yinghai:
Yes, it's accepted. You can upstream that. I think it's fairly reasonable change.

kimishpatel:
@yinghai, I merged all logging to one logger. You might wanna vet the recent changes. For whatever reason though, with these changes file has some binary dump at the beginning and I could not quite figure it out why, so leaving it as is for now. Will fix it later.

PR #125, opened 2019-09-25
Switch to nearbyint instead of using round, which costs performance.

kimishpatel:
Performance loss comes from the fact that llvm does a function call to 'roundf' whereas replacing that with nearbyint just replaces the call with a couple of scale and round instructions.


kimishpatel:
Thanks for quick approval @yinghai. Yes accuracy wise it stays the same.

PR #126, opened 2019-09-25
Use tensorization for avx2 as well.

kimishpatel:
Use avx2 based tensorization to enable perf improvement on avx2 only machines. Gets on par with fbgemm but on VMs variation prevents strong conclusion.


kimishpatel:
dmlc/tvm has reverted the changes on which this PR depends. Even though facebookexperimental/tvm head is behind it and it should work fine, but I would just suggest we dont merge this yet.

PR #129, opened 2019-10-17
Use the C++ version of intrinsic

llyfacebook:


llyfacebook:
> LGTM. Just two things:
> 
> * do we have test cases for this?
> * can we remove the python version?

1) Yes, the existing quantize linear test will trigger the AVX2 intrinsic. For AVX512 intrinsic, I can only test it locally, since we have no machine supporting that.
2) Hmm, want to reuse the python schedule for autotvm, since we have no c++ version autotvm.

yinghai:
For 2, I mean we can remove it from the trunk, to reduce confusion. You can still use it for tuning at your branch.

Fix the CI and this is good to go.

llyfacebook:
> For 2, I mean we can remove it from the trunk, to reduce confusion. You can still use it for tuning at your branch.
> 
> Fix the CI and this is good to go.

Sure, I will remove all together in another PR, since other ops are still in Python.

PR #143, opened 2019-12-03
CUDA support

ilia-cher:
Add support for CUDA in pytorch/tvm

### DIFF
--- a/torch_tvm/compiler.cpp
+++ b/torch_tvm/compiler.cpp
@@ -1,7 +1,9 @@
 #include "compiler.h"
 #include "operators.h"
+#include "register.h"
 
 #include <ATen/DLConvertor.h>
+#include <torch/csrc/autograd/record_function.h>
 #include <torch/csrc/jit/constants.h>
 #include <torch/csrc/jit/interpreter.h>
 #include <limits>
@@ -13,6 +15,7 @@
 using namespace torch::jit;
 
 using torch_tvm::utils::DLManagedTensorPtr;
+using tvm::runtime::DeviceAPI;
 
 namespace {
 std::vector<DLManagedTensorPtr> set_input(
@@ -34,15 +37,20 @@ std::vector<DLManagedTensorPtr> set_input(
     //auto tensor = ivalue.toTensor().to(at::kFloat);
     auto tensor = ivalue.toTensor();
     DLManagedTensor* dl_tensor;
-    if (tensor.is_contiguous() &&
-        torch_tvm::utils::isAligned(tensor.data_ptr(),
-          tvm::runtime::kAllocAlignment)) {
+    if (tensor.is_contiguous() && (
+        tvm::is_training() || torch_tvm::utils::isAligned(tensor.data_ptr(),
+            tvm::runtime::kAllocAlignment))) {
       dl_tensor = at::toDLPack(tensor);
     } else {
-      dl_tensor =
-        torch_tvm::utils::allocAndCopyData(tensor);
-      input_tensors.emplace_back(
-          dl_tensor);
+      if (tvm::is_training()) {
+        auto contig_tensor = tensor.contiguous();
+        dl_tensor = at::toDLPack(contig_tensor);
+      } else {
+        dl_tensor =
+          torch_tvm::utils::allocAndCopyData(tensor);
+        input_tensors.emplace_back(
+            dl_tensor);
+      }
     }
     cache.set_input(graph_input.tvm_var_name,
         tvm::runtime::NDArray::FromDLPack(dl_tensor));
@@ -52,7 +60,17 @@ std::vector<DLManagedTensorPtr> set_input(
 
 DLManagedTensorPtr createParamTensor(const IValue& param_val) {
   auto tensor = param_val.toTensor();
-  auto dl_tensor = torch_tvm::utils::allocAndCopyData(tensor);
+  DLManagedTensor* dl_tensor = nullptr;
+  if (tvm::is_training()) {
+    if (tensor.is_contiguous()) {
+      dl_tensor = at::toDLPack(tensor);
+    } else {
+      auto contig_tensor = tensor.contiguous();
+      dl_tensor = at::toDLPack(contig_tensor);
+    }
+  } else {
+    dl_tensor = torch_tvm::utils::allocAndCopyData(tensor);
+  }
   return DLManagedTensorPtr(dl_tensor);
 }
 
@@ -60,6 +78,33 @@ tvm::relay::Constant createParamConstant(
     const DLManagedTensorPtr& dl_tensor_ptr) {
   auto nd_array = tvm::runtime::NDArray::FromDLPack(dl_tensor_ptr.get());
   return tvm::relay::ConstantNode::make(nd_array);
+}
+
+template <class T, int size>
+tvm::relay::Expr doubleNode(double d, TVMContext ctx) {
+  TORCH_CHECK(size == 16 || size == 32);
+  std::string type;
+  if (size == 16) {
+    type = "float16";
+  } else {
+    type = "float32";
+  }
+
+  auto x = tvm::runtime::NDArray::Empty(
+      {}, tvm::runtime::String2TVMType(type), ctx);
+  TORCH_CHECK(d <= std::numeric_limits<T>::max());
+  TORCH_CHECK(d >= std::numeric_limits<T>::lowest());
+  T f = static_cast<T>(d);
+
+  DeviceAPI::Get(ctx)->CopyDataFromTo(
+      &f, 0,
+      x->data, 0,
+      size,
+      cpuContext(),
+      ctx,
+      DLDataType{kDLFloat, size, 1}, nullptr);
+
+  return tvm::relay::ConstantNode::make(x);
 }
 
 } // namespace
@@ -91,6 +136,7 @@ tvm::Map<std::string, tvm::relay::Constant>
 
 tvm::relay::DataType scalarTypeToTVMType(at::ScalarType pt_type) {
   static const std::unordered_map<at::ScalarType, tvm::relay::DataType> type_mapping = {
+    {at::ScalarType::Half, ::tvm::Float(16)},
     {at::ScalarType::Float, ::tvm::Float(32)},
     {at::ScalarType::Double, ::tvm::Float(64)},
     {at::ScalarType::Int, ::tvm::Int(32)},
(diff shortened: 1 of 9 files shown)


torch_tvm/compiler.cpp
```diff
@@ -34,6 +36,14 @@ std::vector<DLManagedTensorPtr> set_input(
     //auto tensor = ivalue.toTensor().to(at::kFloat);
     auto tensor = ivalue.toTensor();
     DLManagedTensor* dl_tensor;
+#ifdef TVM_TRAINING
```
bwasti: can this logic be merged below and throw a runtime error the tensor is unaligned/non-contiguous and `TVM_TRAINING` is defined?

torch_tvm/compiler.cpp
```diff
@@ -52,7 +63,17 @@ std::vector<DLManagedTensorPtr> set_input(
 
 DLManagedTensorPtr createParamTensor(const IValue& param_val) {
   auto tensor = param_val.toTensor();
-  auto dl_tensor = torch_tvm::utils::allocAndCopyData(tensor);
+  DLManagedTensor* dl_tensor = nullptr;
+#ifdef TVM_TRAINING
```
bwasti: I wonder if this would be better suited as a runtime parameter? recompiling just to train seems a bit odd. Ideally we'd merge the code here anyway, right? (fixing up allocAndCopyData somehow)

torch_tvm/compiler.cpp
```diff
@@ -101,6 +122,7 @@ tvm::relay::DataType scalarTypeToTVMType(at::ScalarType pt_type) {
     {at::ScalarType::QInt8, ::tvm::Int(8)},
     {at::ScalarType::QUInt8, ::tvm::UInt(8)},
     {at::ScalarType::QInt32, ::tvm::Int(32)},
+    {at::ScalarType::Half, ::tvm::Float(16)},
```
bwasti: nit: maybe pop this up at the top just so it's closer to the other Floats?

torch_tvm/compiler.cpp
```diff
@@ -160,30 +179,64 @@ tvm::relay::Var TVMCompiler::convertToRelay(Value* val, TVMContext ctx) {
 tvm::relay::Expr TVMCompiler::convertToRelay(
     const IValue& val,
     TVMContext ctx) {
+  TVMContext cpu_ctx;
+  cpu_ctx.device_type = kDLCPU;
+  cpu_ctx.device_id = 0;
+
   // All doubles are converted to floats
```
bwasti: nit: update this comment? training only done in fp16

torch_tvm/compiler.cpp
```diff
@@ -160,30 +179,64 @@ tvm::relay::Var TVMCompiler::convertToRelay(Value* val, TVMContext ctx) {
 tvm::relay::Expr TVMCompiler::convertToRelay(
     const IValue& val,
     TVMContext ctx) {
+  TVMContext cpu_ctx;
```
bwasti: maybe this could be a static variable? or even one stored in TVMCompiler?

torch_tvm/compiler.cpp
```diff
@@ -343,21 +419,26 @@ TVMCompiler::TVMCompiler(
...
   if (device_type_ == "gpu") {
     ctx_.device_type = kDLGPU;
   } else {
     ctx_.device_type = kDLCPU;
   }
-  ctx_.device_id = 0;
+  ctx_.device_id = device_id_;
   subgraph_ = node->g(attr::Subgraph);
+  handle_str_ = getTVMCompilerHandle(subgraph_);
+  fallback_state_ = std::unique_ptr<torch::jit::InterpreterState>(
```
bwasti: nit: rename `fallback_interpreter_`

torch_tvm/compiler.cpp
```diff
@@ -368,27 +449,50 @@ void TVMCompiler::run(Stack& stack) {
...
+      } else if (kv.second.isDouble()) {
+        kv.first->setType(FloatType::get());
       } else {
-        AT_CHECK(
-            0,
-            "Cannot handle this type yet ",
-            kv.second,
-            "\nGraph:\n",
-            *subgraph_);
+        std::cout << "Cannot handle this type yet "
```
bwasti: let's use a logging utility (there are a couple in PT) rather than std::cout

torch_tvm/compiler.h
```diff
@@ -67,13 +69,20 @@ struct TVMCompiler {
...
   tvm::runtime::Module build_mod_;
   DebugLogger debug_logger_;
   // Not used in OSS
   at::Tensor activation_buffer_;
   std::vector<int64_t> activation_buffer_shape_;
   int64_t max_activation_buffer_size_{0};
 
+  std::string handle_str_;
+  std::unique_ptr<torch::jit::InterpreterState> fallback_state_;
+  std::unordered_set<torch::jit::CompleteArgumentSpec> bad_specs_;
```
bwasti: nit: add a comment explaining the purpose of this param -> to cache uncompiled-able specs

torch_tvm/operators.cpp
```diff
@@ -35,6 +36,13 @@ const tvm::relay::TensorTypeNode getExprType(const tvm::relay::Expr& input) {
   return *checked_type;
 }
 
+TVMContext cpuContext() {
```
bwasti: looks like this is exactly the "static" cpu_ctx I was looking for above :) let's swap that instance with this

torch_tvm/operators.cpp
```diff
@@ -321,7 +338,7 @@ RegisterTVMOperator::RegisterTVMOperator(std::vector<TVMOpMap> ops) {
         // NB: We assume all relay ops are pure
         auto options = c10::OperatorOptions();
         options.setAliasAnalysis(AliasAnalysisKind::PURE_FUNCTION);
-	      // TODO: Pass in operator options somehow
+             // TODO: Pass in operator options somehow
```
bwasti: nit: formatting is strange here

torch_tvm/operators.cpp
```diff
@@ -746,7 +785,7 @@ RegisterTVMOperator reg({
        attrs->newshape = relayToArray<tvm::Integer>(inputs[1]);
        TORCH_INTERNAL_ASSERT(attrs->newshape.size() > 0);
        if (static_cast<int64_t>(attrs->newshape[0]) != -1) {
-	       std::cerr << "WARNING: reshape with -1 as the first value has known incompatibility with PyTorch semantics.\n";
+         std::cerr << "WARNING: reshape with -1 as the first value has known incompatibility with PyTorch semantics.\n";
```
bwasti: when you change the std::cout above to the logging framework, mind changing this one as well?

ilia-cher: will upd

torch_tvm/operators.cpp
ilia-cher: will check

torch_tvm/operators.cpp
ilia-cher: sounds good

torch_tvm/compiler.h
ilia-cher: yep

torch_tvm/compiler.cpp
ilia-cher: I'll check

torch_tvm/compiler.cpp
ilia-cher: yep

torch_tvm/compiler.cpp
ilia-cher: we have some issues with static initialization unfortunately that make it hard to use a runtime parameter

torch_tvm/compiler.cpp
yinghai: Actually I don't understand why this is a static flag either and why training requires this. Is this more of a CPU vs CUDA thing?

ilia-cher: I updated it to a runtime flag in the latest update

torch_tvm/compiler.cpp
```diff
@@ -34,15 +37,20 @@ std::vector<DLManagedTensorPtr> set_input(
     //auto tensor = ivalue.toTensor().to(at::kFloat);
     auto tensor = ivalue.toTensor();
     DLManagedTensor* dl_tensor;
-    if (tensor.is_contiguous() &&
-        torch_tvm::utils::isAligned(tensor.data_ptr(),
-          tvm::runtime::kAllocAlignment)) {
+    if tensor.is_contiguous() && (
+        is_training() || torch_tvm::utils::isAligned(tensor.data_ptr(),
```
yinghai: Could you add comment why during training, we require the tensor to be contiguous?

ilia-cher: we always require tensor to be contiguous (or make it contiguous)

yinghai: OK. Sorry, just comment on why we need special handling for training?

ilia-cher: sg

kimishpatel: Previously we were checking not only for contiguity but also for 64B alignment. If either fails then we do copy. The new condition seems to assume that in training data is always aligned to 64B?

kimishpatel: Or in training we never want to make a copy?

torch_tvm/compiler.cpp
```diff
@@ -131,9 +171,6 @@ tvm::relay::Var TVMCompiler::convertToRelay(Value* val, TVMContext ctx) {
     TORCH_INTERNAL_ASSERT(pt_t);
     auto optional_device_type = pt_t->device();
     TORCH_INTERNAL_ASSERT(optional_device_type);
-    auto device_type = optional_device_type.value();
-    AT_CHECK(device_type == at::DeviceType::CPU,
-      "Expected CPU device type but got:", device_type);
```
kimishpatel: Fix the comment above.

torch_tvm/compiler.cpp
ilia-cher: we dont' check for alignment here

kimishpatel: So how are we adding cuda support here? By being able to specify context for devices other than cpu?

torch_tvm/compiler.cpp
```diff
@@ -428,14 +532,15 @@ void TVMCompiler::run(Stack& stack) {
     // being partitioned in non-divisible factors it will partition
     // it in divisible part and tail that is not divible.
     // This help performance.
-    build_config->partition_const_loop = true;
+    build_config->partition_const_loop = false;
```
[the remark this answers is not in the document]
ilia-cher: this is an old, I have already reverted it .....

ilia-cher:
@kimishpatel  yes that's the whole point of this PR to be able to use context other than cpu

yinghai:
We need to update CMakefile? https://github.com/pytorch/tvm/blob/a34c88da857974d8d40ae0858b5bbb8b6513bd9d/CMakeLists.txt#L10

ilia-cher:
addressed comment

